PR #494, opened 2015-01-23
Fix Python 3 detection

MultiplyByZer01:
This fixes the code that detects whether it is running on Python 3. This prevents errors on Python 3.2 or other older versions of Python 3.


nicolargo:
@asergi What do you think ?

asergi:
We do not support Python >= 3.0, < 3.3.

Python < 3.3 do not support unicode literal.
Python 3.2 is no longer maintained and the userbase, if any, is very small, however.
Also, 3.0 and 3.1 do not exist anymore.

Reason for closing: Not a bug.

PR #496, opened 2015-01-24
Improve rendering of the Web UI loading page on small devices

nclsHart:
Changes : 
* use em instead of px
* the dots animation is now visible even on small devices
* add link to provide initial source of the animation


nicolargo:
Done. Please include upstream (DEVELOP) change before pushing/pull request.

PR #517, opened 2015-02-25
Avoid crashing if LC_ALL is not defined by user

jcberthon:
When using SSH, my LC_ALL environment variable is not set (sshd configuration).
This seems to be a default configuration on Raspbian (equiv Debian 7.8) and Ubuntu 14.04.2.

The problem is that when launching glances via SSH in this manner, the program crash with a locale.Error exception line 98 in `__init.py__`. I'm no python specialist but as far as I understood the python 2.7 documentation, the statement `locale.setlocale(locale.LC_ALL, '')` try to reset the locale for the python program to the user settings, which in my case are not defined. And I suspect that's the root of the problem.

So I propose a patch (commit `aadb49f`) where I do a try/except and simply pass the error, as anyway from the documentation the statement `locale.setlocale(locale.LC_ALL, '')` should always work.

*Note: after that change, I can successfully run glances via SSH on my Raspberry Pi running Raspbian. See screenshot:*
glances on raspberry pi

### DIFF
--- a/glances/__init__.py
+++ b/glances/__init__.py
@@ -95,7 +95,13 @@ def main():
     Run it...
     """
     # Setup translations
-    locale.setlocale(locale.LC_ALL, '')
+    try:
+        locale.setlocale(locale.LC_ALL, '')
+    except locale.Error:
+        # Setting LC_ALL to '' should not generate an error unless LC_ALL is not
+        # defined in the user environment, which can be the case when used via SSH.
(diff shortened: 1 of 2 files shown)


nicolargo:
Thk for the pull, i am going to merge it to the DEVELOP branch (not the master one). The patch will be available for the next Glances release (Glances 2.4).

PR #527, opened 2015-03-17
make doco a little bit more discoverable

DrHyde:
A teeny-tiny patch to make it easier to find the --theme-white option. I expect I'm not the only one who searched for a keyword instead of reading the whole man-page.


nicolargo:
Merged in the DEVELOP branch.
Thks for the contrib.

PR #534, opened 2015-04-01
don't crash on older kernels

molobrakos:
catch exception NotImplementedError thrown by psutil on older kernels: NotImplementedError: couldn't find /proc/1/io (kernel too old?)


nicolargo:
Merged in the DEVELOP branch.
Thk for the tips.

PR #535, opened 2015-04-01
Add docker plugin on angularjs web ui

nclsHart:
Preview : 

capture d ecran - 02042015 - 01 19 37


nicolargo:
One point @nclsHart : The plugin is displayed even if there is no running container... Do ou think that it is possible to optimize this ?

nclsHart:
Oops. I'll fix this issue asap.

PR #536, opened 2015-04-04
[Web UI] Display the Docker plugin only if there is a running container

nclsHart:
Fix issue reported in #535.


nicolargo:
Perfect !

PR #550, opened 2015-05-01
[Web UI] Processlist sort + monitor plugin

nclsHart:
Main changes : 

* fix docker container cpu not rounded
* fix process list sorting
* add monitor plugin

Monitor plugin preview : 
glances-monitor


nicolargo:
Nice :)

PR #559, opened 2015-05-06
Fix MANIFEST for the web UI

notFloran:
Add HTML and images files for the web UI in the MANIFEST


nicolargo:
Hi @notFloran thanks for the 3 contribs !

PR #592, opened 2015-05-09
[Web UI] Load limit take Core number into account

### LINKED ISSUE #587: [Web UI] Load should take core number into account
The load limit depends of the number of Core.
https://github.com/nicolargo/glances/blob/develop/glances/plugins/glances_load.py#L127

Web UI:
selection_359

Curses:
selection_360

notFloran:
Fix #587 

The current current way of managing limits/alerts do not take into account the "maximum" in calculations.
So I refactored the method to be closer to what is done in python https://github.com/nicolargo/glances/blob/develop/glances/plugins/glances_plugin.py#L352

### DIFF
--- a/glances/outputs/static/html/plugins/cpu.html
+++ b/glances/outputs/static/html/plugins/cpu.html
@@ -5,13 +5,13 @@
     </div>
     <div class="table-row">
         <div class="table-cell text-left">user:</div>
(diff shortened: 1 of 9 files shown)


asergi:
Please rebase to make this applicable.

notFloran:
Done !

PR #630, opened 2015-07-08
Support InfluxDB 0.9.x

emidln:
This adds support for the JSON payload schema used in 0.9.x. This
also adds support for the tags feature available in 0.9.x via the
config option tags. Tags is a comma separated list of colon
separated key:value pairs. An example:

```
[influxdb]
tags = foo:bar,spam:eggs,region:us-east-1
```


nicolargo:
Perfect !

Thk Brandon !

PR #634, opened 2015-07-16
Fix UTF-8 characters being output on latin1 terminals

h3xx:
This fix translates the characters to an appropriate encoding for the
terminal. "Â°C" (A-caret, degrees, C) was being output on latin1 terminals, which is a bug
since the LC_CTYPE locale setting should tell the program whether it's
okay to output Unicode.

This follows the recommendation from the official curses library
documentation: https://docs.python.org/2/library/curses.html

Steps to reproduce the issue:

1. Install pysensors, enable any kernel modules necessary to support it
2. Run `LANG=en_US rxvt -e glances`
3. Look in SENSORS section

The corrupt Unicode character sequence will appear.

### DIFF
--- a/glances/outputs/glances_curses.py
+++ b/glances/outputs/glances_curses.py
@@ -22,6 +22,7 @@
 # Import system lib
 import re
 import sys
+import locale
 
 # Import Glances lib
 from glances.core.glances_globals import is_mac, is_windows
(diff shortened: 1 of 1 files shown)


nicolargo:
Thanks for the contrib. Please do not pull on the master branch but only in the develop one.

Read: https://github.com/nicolargo/glances/wiki

h3xx:
I came up with a less CPU-intensive way of fixing it.

PR #636, opened 2015-07-20
unittest: start server in a modified environment.

eroullit:
This changes allows the user to run the unittests properly in
a virtualenv setup. Once the venv activated, /usr/bin/env
will point the python executable within the venv instead of
the system one.


asergi:
Thanks.

eroullit:
Glad I could help.

PR #653, opened 2015-08-20
[Web UI] Warn user if the server is down/disconnected

### LINKED ISSUE #564: [Web UI] No way to know if the server is not connected
From the Web UI, there is no way to know if the server is down/disconnect.

The Glances WebUI should display a message to warn user that the displayed stats are not up to date.

Perhaps best way to process is display the same message than the Curses client/server:

selection_355

nclsHart:
glances-disconnected

Fix #564.


nicolargo:
Thk Nicolas ! :+1:

PR #656, opened 2015-08-22
[Web UI] Big refactoring

notFloran:
Big refactoring of the Web UI with the aim to have something closer to what is done in Python code with the plugin system.

No more logic in the controller or views, all the logic is in plugins (*glances/outputs/static/js/services/plugins*).
All the API calls are made in the GlancesStats service which is also responsible to set data (stats, views, limits).

The web UI now uses views data when is possible instead of computing limits to get alert.


nicolargo:
Thk Floran ! A good work. Hope it will help for future WebUI enhancement.

PR #664, opened 2015-09-05
[Web UI] Quicklook plugin

nclsHart:
Preview : 

selection_006

selection_007

Other changes : 

* The Web UI is now using the same colors as in the curse interface
* The Web UI is now using a non standard 24 columns layout in order to have more precision for plugin placement


nicolargo:
Nice pull !!!

Issue #670, opened 2015-09-10
FreeBSD 10.1 Crashes


hank:
No matter what I do, glances crashes in freebsd 10.1 for me with python2.7 from pkg and ports.  Here's a core backtrace:
```
#0  0x0000000800adefbf in PyTuple_Pack () from /usr/local/lib/libpython2.7.so.1
#1  0x000000080748220c in gateways () from /usr/local/lib/python2.7/site-packages/netifaces.so
#2  0x0000000800b269b1 in PyEval_EvalFrameEx () from /usr/local/lib/libpython2.7.so.1
#3  0x0000000800b22bc4 in PyEval_EvalCodeEx () from /usr/local/lib/libpython2.7.so.1
#4  0x0000000800ab388c in PyFunction_SetClosure () from /usr/local/lib/libpython2.7.so.1
#5  0x0000000800a8fef4 in PyObject_Call () from /usr/local/lib/libpython2.7.so.1
#6  0x0000000800b261e9 in PyEval_EvalFrameEx () from /usr/local/lib/libpython2.7.so.1
#7  0x0000000800b22bc4 in PyEval_EvalCodeEx () from /usr/local/lib/libpython2.7.so.1
#8  0x0000000800b2a439 in _PyEval_SliceIndex () from /usr/local/lib/libpython2.7.so.1
#9  0x0000000800b25f2b in PyEval_EvalFrameEx () from /usr/local/lib/libpython2.7.so.1
#10 0x0000000800b2a3d2 in _PyEval_SliceIndex () from /usr/local/lib/libpython2.7.so.1
#11 0x0000000800b25f2b in PyEval_EvalFrameEx () from /usr/local/lib/libpython2.7.so.1
#12 0x0000000800b22bc4 in PyEval_EvalCodeEx () from /usr/local/lib/libpython2.7.so.1
#13 0x0000000800ab388c in PyFunction_SetClosure () from /usr/local/lib/libpython2.7.so.1
#14 0x0000000800a8fef4 in PyObject_Call () from /usr/local/lib/libpython2.7.so.1
#15 0x0000000800a9bee3 in PyMethod_New () from /usr/local/lib/libpython2.7.so.1
#16 0x0000000800a8fef4 in PyObject_Call () from /usr/local/lib/libpython2.7.so.1
#17 0x0000000800ae687d in _PyObject_SlotCompare () from /usr/local/lib/libpython2.7.so.1
[...]

PR #671, opened 2015-09-10
FreeBSD crash workaround and psutil error check

hank:
This fixes a problem I was having with psutil after running for days or weeks (no more glances exceptions thrown!).  It also fixes the FreeBSD segfault from netifaces.gateways in Issue #670.


nicolargo:
Thanks for the patch @hank .
Patch will not be merged in the master branch but in the develop one (as explained in the Wiki: https://github.com/nicolargo/glances/wiki).
I will pick your code and include it in the develop branch by myself.

nicolargo:
REJECTED: Already solved into the develop branch

PR #683, opened 2015-09-20
Change percentage bars to use an explicit exception for value checks

kevinlondon:
The percentage bars will now use an AssertionError directly if the provided values are outside of the expected range (0-100 inclusive). 

I was using [Bandit](https://github.com/openstack/bandit) to check for potential vulnerabilities in a few popular Python projects and this came up as a warning:

```
>> Issue: Use of assert detected. The enclosed code will be removed when compiling to optimised byte code.
    Severity: Low   Confidence: High
    Location: glances/glances/outputs/glances_bars.py:72
 70      @percent.setter
 71      def percent(self, value):
 72          assert value >= 0
 73          assert value <= 100
 74          self.__percent = value
```

As per the warning, `assert` statements are stripped out if compiling with optimized bytecode (-O option) in Python. Here's a [StackOverflow question](http://stackoverflow.com/questions/2830358/what-are-the-implications-of-running-python-with-the-optimize-flag) with a little discussion and the original [Python docs](https://docs.python.org/2/reference/simple_stmts.html#the-assert-statement). 

Oh, as a quick postscript, these tests run fine on OSX, so perhaps the Linux checks could be made to include OSX when running the tests?

### DIFF
--- a/glances/outputs/glances_bars.py
+++ b/glances/outputs/glances_bars.py
@@ -69,8 +69,9 @@ class Bar(object):
 
     @percent.setter
     def percent(self, value):
-        assert value >= 0
-        assert value <= 100
+        if value < 0 or value > 100:
+            raise AssertionError('The percent must be between 0 and 100.')
+
         self.__percent = value
 
     @property
--- a/unitest.py
+++ b/unitest.py
@@ -24,6 +24,7 @@ import sys
 import time
 import unittest
 
+from glances.outputs.glances_bars import Bar
 from glances.core.glances_globals import (
     appname,
     is_linux,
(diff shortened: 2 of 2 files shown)


asergi:
Hi @kevinlondon,

Thanks for your contribution. I manually merged the PR into develop with commit https://github.com/nicolargo/glances/commit/2b1b80c0323f89baabc7b71fcb7db549be9cddfc.

I guess those `assert` statements were added for debugging purposes and then left there.
About the tests, I will have a look later.

Thanks again for your contribution.

asergi:
Tests unblocked: https://github.com/nicolargo/glances/commit/837193ef7bdaf2b37d176cd750235ab0008c2e12.

PR #704, opened 2015-10-19
[WebUI] add "pointer" cursor for sortable columns

notFloran:
Add a "pointer" cursor to indicate which columns are sortable


nicolargo:
Nice !

PR #706, opened 2015-10-20
[Web UI] Fix no nice decoration in processlist

nclsHart:
Before : 

selection_003

After : 

selection_004


nicolargo:
Thk Nicolas

Issue #724, opened 2015-11-24
[WebUI] Make web page title configurable


telethonic:
a minor feature request: can you make it so I can set the web page title in the config file?

this way, when I have multiple tabs open on multiple servers, I can see which is which
(now they all say "Glances")

thanks :)

nicolargo:
Another solution is to fill the title with the client IP address / name instead of the default Glances text. What do you think ?

telethonic:
That would work. 

But in my case (running glances on Amazon AWS instances) the names and IP addresses are not always very informative (they're typically "ip-<internal ip address>"). That said, I can of course change the hostname :)

So whatever is easiest.

nicolargo:
Ok.

@notFloran Can you have a look to implement this feature into the WebUI ?

notFloran:
@nclsHart I made a PR for that :)

nicolargo:
@telethonic Implemented in the latest commit (see issue #726). Can you test the develop branch ?

PR #726, opened 2015-11-25
[WebUI] Use hostname for title

notFloran:
See #724

Result : 
```<hostname> - Glances```

capture d ecran 2015-11-25 a 13 29 56


nicolargo:
Thk @notFloran

Issue #68, opened 2012-09-04
Colors


samirahmed:
Is there anyway to customize the colors,  I use a custom color setting and it seems to make the okay green illegible.

PR #735, opened 2015-12-01
Add option to disable ANSI background colours

okdana:
I've just found Glances, and i really like it, but i've had the same issue described in #68 — the ANSI background colours (especially green) clash with my terminal colour scheme, making the CPU percentages unreadable.

This PR adds an option `--disable-bg` which turns off the ANSI background colours: white-on-green becomes standard green, white-on-magenta becomes standard magenta, &c.

I modelled the new option after the `--disable-bold` one, whose variable seems to be stored backwards from what i'd expect (`args.disable_bold == True` actually means bold is *not* disabled). I don't get why it's like that, but i copied the same behaviour for the new option.

### DIFF
--- a/docs/glances-doc.rst
+++ b/docs/glances-doc.rst
@@ -168,6 +168,7 @@ Command-Line Options
                         sensors modules (py3sensors needed)
   --disable-process     disable process module
   --disable-log         disable log module
+  --disable-bg          disable background colors in the terminal
   --disable-bold        disable bold mode in the terminal
   --enable-process-extended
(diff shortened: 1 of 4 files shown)


nicolargo:
Hi @okdana . Thanks for your pull request. Infortunately, we only accept pull on the develop branch. Have a look on:  https://github.com/nicolargo/glances/blob/master/CONTRIBUTING.md

okdana:
I'm sorry, i didn't even notice. I've created a new PR #738, closing this one.

PR #738, opened 2015-12-02
Add option to disable ANSI background colours (2)

okdana:
This replaces PR #735, which was:

> I've just found Glances, and i really like it, but i've had the same issue described in #68 — the ANSI background colours (especially green) clash with my terminal colour scheme, making the CPU percentages unreadable.

> This PR adds an option `--disable-bg` which turns off the ANSI background colours: white-on-green becomes standard green, white-on-magenta becomes standard magenta, &c.

> I modelled the new option after the `--disable-bold` one, whose variable seems to be stored backwards from what i'd expect (`args.disable_bold == True` actually means bold is *not* disabled). I don't get why it's like that, but i copied the same behaviour for the new option.

### DIFF
--- a/docs/glances-doc.rst
+++ b/docs/glances-doc.rst
@@ -169,6 +169,7 @@ Command-Line Options
                         sensors modules (py3sensors needed)
   --disable-process     disable process module
   --disable-log         disable log module
+  --disable-bg          disable background colors in the terminal
   --disable-bold        disable bold mode in the terminal
   --enable-process-extended
                         enable extended stats on top process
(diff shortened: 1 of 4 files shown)


parkerlreed:
Is there a way to turn off colors completely?

nicolargo:
@parkerlreed 

you can try this:

    TERM=xterm-mono glances

gtimchenko:
Since --disable-bg is implemented as an option - is there a way to apply this in the configuration file?
disable-bg=true doesn't work in [global] and [outputs] both

PR #747, opened 2015-12-08
Fix warning on code-block directives

martin-damien:
An argument is mandatory for the `code-block` directives.


nicolargo:
Thanks !

PR #756, opened 2015-12-14
Fix hide interface config not take into account in web ui

### LINKED ISSUE #719: [Web UI] Hide interface configuration tag not take into account in the Web UI
As said in the issue #714 , it seems that -w mode does not honor settings in .conf

    [diskio]
    hide=ram0,ram1,ram2,ram3,ram4,ram5,ram6,ram7,ram8,ram9,ram10,ram11,ram12,ram13,ram14,ram15

nclsHart:
Fix #719


nicolargo:
Thks (again) @nclsHart !

PR #783, opened 2016-01-22
Fix #768

### LINKED ISSUE #768: '--tree' breaks process filter on Debian 8
I saw similar issue with OSX/FreeBSD/Windows #542 where it was claimed that this was an bug exclusive to those operating systems. However I have a similar bug on a dedicated machine running __Debian 8.2__.

Running `glances --tree` and trying to filter the processes leads to glances being stuck at 100% CPU on one core.

Sometimes the prompt (with the filter) stays open, sometimes it manages to disappear before glances freezes.

I am connecting from a windows machine, using cygwin+tmux and tmux inside the root. But I don't think this is the reason for the freeze. Some information about the root:

```sh
python --version                                                                 
Python 2.7.9

glances --version                                                                
Glances v2.5.1 with psutil v3.3.0

lsb_release -a                                                                   
No LSB modules are available.
Distributor ID: Debian
Description:    Debian GNU/Linux 8.2 (jessie)
Release:        8.2
Codename:       jessie

uname -a                                                                         
Linux hetzner-root 3.16.0-4-amd64 #1 SMP Debian 3.16.7-ckt20-1+deb8u2 (2016-01-02) x86_64 GNU/Linux
```

desbma:
Fix issue #768

### DIFF
--- a/glances/processes.py
+++ b/glances/processes.py
@@ -177,7 +177,7 @@ class ProcessTreeNode(object):
             nodes_to_search.extend(current_node.children)
 
     @staticmethod
-    def build_tree(process_dict, sort_key, sort_reverse, hide_kernel_threads):
+    def build_tree(process_dict, sort_key, sort_reverse, hide_kernel_threads, excluded_processes):
         """Build a process tree using using parent/child relationships.
 
         Return the tree root node.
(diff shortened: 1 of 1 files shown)


nicolargo:
Thks for the fix @desbma

PR #785, opened 2016-01-25
[WebUI] Dont wait before first loading

notFloran:
With this fix, the WebUI makes the first loading of data then refresh the data every x seconds.


nicolargo:
Perfect !

PR #786, opened 2016-01-25
Fix #546

### LINKED ISSUE #546: Binary highlighting in TASKS-list is incorrect when path contains whitespaces
On OSX the green highlighting of the binary in the task list does not properly handle white spaces in the path e.g.

> /Applications/**Google** Chrome.app/Contents/MacOS/Google Chrome

expected behaviour would be

> /Applications/Google Chrome.app/Contents/MacOS/**Google Chrome**

(bold should be highlighted)

it's really not high priority just something I noticed.

desbma:
To test, run in a Bash shell (Linux):

    echo "void main() { pause(); }" > /tmp/s.c
    gcc /tmp/s.c -o '/tmp/a b'
    '/tmp/a b' a b c d &

Then start glances, you should see '/tmp/**a b** a b c d' in process list.

### DIFF
--- a/glances/plugins/glances_processlist.py
+++ b/glances/plugins/glances_processlist.py
@@ -48,12 +48,12 @@ def split_process_cmdline(process):
     basename = process['name']
     arguments = None
     if process['exe'] and process['cmdline'].startswith(process['exe']):
-      dn, bn = os.path.split(process['exe'])
-      dirname = dn or None
-      basename = bn or process['name']
-      arguments = process['cmdline'][len(process['exe']):].lstrip()
+        dn, bn = os.path.split(process['exe'])
+        dirname = dn or None
+        basename = bn or process['name']
+        arguments = process['cmdline'][len(process['exe']):].lstrip()
     elif process['cmdline'].startswith(process['name']):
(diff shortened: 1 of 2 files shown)


nicolargo:
Tested and approved !
Thanks @desbma

asergi:
I got several regressions after applying this patch. Here's some examles:
- `-zsh` -> `zsh`: the leading dash is stripped (wrong); the process is not even highlighted
- `/bin/sh /usr/bin/startx` -> `startx`
- `/bin/sh -c i3bar --bar_id=bar-0 --socket="/tmp/1000-runtime-dir/i3/ipc-socket.1098"` -> `sh`
- some long command lines are displayed correctly while others aren't:

Before: long-cmd-before
After: long-cmd-after

nicolargo:
Mmmhhh...  Can you revert the merge? I am far away from computer today.

desbma:
Since it's in the develop branch, I think we should fix remaining issues instead of reverting.
I'll have a look at @asergi's regressions today.

nicolargo:
Ok @desbma

PR #789, opened 2016-02-07
[WebUI] Use arguments

### LINKED ISSUE #784: Refresh time not take into account in Web Server mode
I was using glances -t 15 in order to force it to update only once every 15 seconds. This worked great, and kept the CPU usage of Glances down! Now I'm trying to use it in web server mode, but -t doesn't seem to do anything. No matter what I do, it always updates once every 3 seconds and is using 13% CPU to do it!

Is there any way to make the webserver mode update less frequently?

Source:  https://groups.google.com/forum/?utm_medium=email&utm_source=footer#!msg/glances-users/CpiI6ISyCsI/PB3WJWs0BwAJ

@nclsHart / @notFloran : can you have a look, refresh time (-t option) does not work as expected in (-w) mode.

notFloran:
The arguments are now taken into account in the WebUI.

Fix #784

### DIFF
--- a/glances/main.py
+++ b/glances/main.py
@@ -247,9 +247,8 @@ Start the client browser (browser mode):\n\
         if args.disable_autodiscover:
             logger.info("Auto discover mode is disabled")
 
-        # In web server mode, defaul refresh time: 5 sec
+        # In web server mode
         if args.webserver:
-            args.time = 5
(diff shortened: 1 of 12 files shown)


nicolargo:
Great job. Look like the WebUI is almost ~ to the Curses one !

PR #798, opened 2016-02-22
Change http://bit.ly to https://bit.ly

quietshu:
Using HTTP will cause a CONNECTION_RESET error:

```
curl -L http://bit.ly/glances | /bin/bash
  % Total    % Received % Xferd  Average Speed   Time    Time     Time  Current
                                 Dload  Upload   Total   Spent    Left  Speed
  0     0    0     0    0     0      0      0 --:--:-- --:--:-- --:--:--     0
curl: (56) Recv failure: Connection reset by peer
```

PR #801, opened 2016-02-24
Update index.html

quietshu:
Fixed bit.ly protocol at http://nicolargo.github.io/glances/, same with #798. :smile:


nicolargo:
The index.html file is generated automaticaly...

PR #803, opened 2016-03-04
[WebUI] hide diskio when no data

notFloran:
Before : 

capture d ecran 2016-03-04 a 10 36 27

After : 

capture d ecran 2016-03-04 a 10 36 13


nicolargo:
Thks !

PR #825, opened 2016-03-27
[Web UI] add cpu name in quicklook plugin

nclsHart:
Preview : 

cpu-name


nicolargo:
Great ! Thk @nclsHart

PR #851, opened 2016-04-17
Fix #849

### LINKED ISSUE #849: Wrong formatting for tree columns
When I run `glances --tree` the tree menu is messed up. It looks like `W/s` and `Command` columns are swapped for some of child processes. If I run glances command multiple times, this always happens for same processes.

screenshot

* Glances: 2.6.1
* PSutil: 4.1.0
* Operating System: ArchLinux

desbma:
Fix issue #849.

It should be more robust when calculating position to add tree decoration.


nicolargo:
Thanks @desbma

Issue #734, opened 2015-11-29
New "Ports scanner" plugin


nicolargo:
The "ports scanner" plugin aims at monitoring local or remote TCP/ICMP ports status. A list of remote host/port is define in the configuration file and the result is displayed in the UI.

Configuration file example:

    [ports]
    port_1_description=Internet connection
    port_1_hostport=www.google.fr:80
    port_2_description=Ping my router
    port_2_hostport=192.168.0.254

Note: if no port is provided, use ICMP instead of TCP.

UI:

    Internet connection    Port unreachable (red)
    Ping my router         10ms (green)

Source:

* https://docs.python.org/2/library/asyncore.html#module-asyncore
* http://stackoverflow.com/questions/1205863/how-can-i-get-non-blocking-socket-connects

For ICMP: all need **root** right !!!
* For ICMP: https://github.com/volftomas/pyicmp/blob/master/ping.py
* For ICMP (bis): https://github.com/samuel/python-ping/blob/master/ping.py
* For ICMP (ter): https://gist.github.com/pklaus/856268
* For ICMP (last but not least): https://bitbucket.org/delroth/python-ping/src/7084c31fa1603a1071c30525ae0431da42ce93cd/ping.py?at=default&fileviewer=file-view-default

Also have a look on http://denilson.sa.nom.br/prettyping/ (for the nice graph idea)

nicolargo:
See and complete the following Gift for this plugin core: https://gist.github.com/nicolargo/d220f67da6d826cc9137

nicolargo:
![selection_033](https://cloud.githubusercontent.com/assets/776747/16008747/ae62daf4-3179-11e6-96bb-6eaf48555978.png)

nicolargo:
Nice proposal from Twitter:

Maxime Carron @mxcarron
@nicolargo it would be nice to get "default gateway” from system config (and keep the ip updated), and have tested by default ;-)

nicolargo:
Implemented in the latest HEAD version (develop branch).

Please test !

nicolargo:
@notFloran or @nclsHart : Can you update the Web UI with this new plugin ? Thanks !

nclsHart:
@nicolargo I'll do it.

nicolargo:
Thks in advance @nclsHart !!!

nclsHart:
@nicolargo There is no shortcut for hiding/displaying this plugin?

nicolargo:
[...]

PR #897, opened 2016-07-20
[Web UI] add ports plugin

nclsHart:
#### Description

ports

#### Resume

* Bug fix: no
* New feature: yes
* Fixed tickets: #734


nclsHart:
@nicolargo It's strange, I get different results (very different) between values displayed by the curse and the web interface.

nicolargo:
Strange... will have a look...

nicolargo:
@nclsHart Do you have the same value when you only ran the Web UI ?

nclsHart:
@nicolargo Values are always lower on web UI even with only one instance running at the same time.

PR #924, opened 2016-09-19
Docker build can't install packages

thebwt:
#### Description

Fixing the dockerfile for your automatic builds, which have been broken for 5 months.
https://hub.docker.com/r/nicolargo/glances/builds/

#### Resume

* Bug fix: yes
* New feature: no
* Fixed tickets: didn't see an issue

Need to initialize the apt repos with `apt-get update`. Since you're maintaining your own install script and not installing cleanly here, we should break the system apart so that we're clearing the apt cache properly.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -8,8 +8,9 @@
 FROM ubuntu
 
 # Install Glances (develop branch)
-RUN apt-get -y install curl && \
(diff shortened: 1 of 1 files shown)


thebwt:
For the record, there are larger problems, the script it uses needs some changes to finish correctly - see nicolargo/glancesautoinstall#11

PR #979, opened 2016-12-16
Implement basic GPU monitoring plugin for NVIDIA cards

kdbanman:
#### Description

This PR is in response to #170.  It is only the plugin implementation, ready for view and test implementation.

I am happy to support and add plugin functionality in the future as well.  Perhaps even AMD card support, should a reasonable AMD GPU monitoring library appear for python.

#### Resume

* Bug fix: no
* New feature: yes
* Fixed tickets: 170

Issue #170, opened 2013-01-28
GPU monitoring (NVidia only)


alexis26:
Bonjour,

Merci pour ce soft que j'utilise et apprécie (presque depuis le début, car je suivais déjà votre blog via rss avant ça...).

Je me demandais si vous pensez qu'il sera (serait ?) possible d'y ajouter les infos GPU.
Vu que de + en +de cartes graphiques ont un proc. dédié, je me disait qu'il serait intéressant d'avoir le même type d'info qu'avec glances... et éventuellement même de voir la décharge de l'un sur l'autre ...

Merci
Alexis

nicolargo:
Best way to proceed is to write a dedicated Python lib with support for:
- Intel
-Nvidia
- ATI

Glances will use this API to retreive GPU stats.

Bedotech:
Any news?

kdbanman:
This feature would make Glances better for sure.  It's already amazing, so this would be icing on an already iced cake 🍰

nicolargo:
Could be possible to do a first version for Nvidia GPU because a Python lib already exist (https://pypi.python.org/pypi/nvidia-ml-py/). For AMD and INTEL, i am looking for existing Python Lib.

kdbanman:
@nicolargo Keen observation - I was just doing that!  I actually have a half-implemented version that uses that library.  The stats collection works, but I just can't quite find all the places to tell Glances about it.

There's the `GlancesPlugin` to override, the webapp angular plugin and view, the REST API, and some others.  I could just `grep -nr "cpu"` and make my gpu plugin fill all the same holes, but do you have a document that lays out the plugin addition process?

nicolargo:
@kdbanman Happy to heard that you have started to work on this enhancement. 

For the moment there is no official documentation / check-list about plugin addition. You can have a look on this pull request which is a good example: https://github.com/nicolargo/glances/pull/911

Or, if you want, you can write a first version of the plugin (only the plugin script) and i will create a branch will all the stuff arround it to test it on your system (i do not have any Nvidia card...).

What do you think ?

kdbanman:
@nicolargo That PR is exactly what I needed.  I can work from there.  3 questions:

1. Do you want me to branch off `develop` like that PR did?
2. If I have questions about that PR specifically, can I ask on that PR's conversation thread?
[...]

PR #982, opened 2016-12-18
Fix incorrectly registered GPU id in gpu plugin

kdbanman:
#### Description

Single line fix to correctly register the gpu plugin key for each device id.  This is a continuation of #170 and #979.

This PR is now tested.  For instance, here is the API output for an **old** card (which does not support the processor utilization query in the Nvidia management library:

1.  No load

```
$ curl http://localhost:61208/api/2/gpu
[{"memory_percent": 0, "processor_percent": null, "gpu_id": 0, "name": "GeForce GTX 560 Ti"}]
```

2. Training a 128 node, single hidden layer recurrent neural net

```
$ curl http://localhost:61208/api/2/gpu
[{"memory_percent": 14, "processor_percent": null, "gpu_id": 0, "name": "GeForce GTX 560 Ti"}]
```

3. Training a 1024 node-per-layer, 3 hidden layer recurrent neural net

```
$ curl http://localhost:61208/api/2/gpu
[{"memory_percent": 98, "processor_percent": null, "gpu_id": 0, "name": "GeForce GTX 560 Ti"}]
```

The important thing to note here is that, despite the card being old and not supporting all of the Nvidia management library API, the gpu plugin does not crash.

Right now I am repeating these tests on a more modern card to make sure that better cards are actually supported.

#### Resume

* Bug fix: yes
* New feature: yes
* Fixed tickets: 170, 979

### DIFF
--- a/glances/plugins/glances_gpu.py
+++ b/glances/plugins/glances_gpu.py
@@ -108,7 +108,7 @@ class Plugin(GlancesPlugin):
 
         for index, device_handle in enumerate(self.device_handles):
             device_stats = {}
-            device_stats['key'] = index
+            device_stats['gpu_id'] = index
             device_stats['name'] = self.get_device_name(device_handle)
             device_stats['memory_percent'] = self.get_memory_percent(device_handle)
             device_stats['processor_percent'] = self.get_processor_percent(device_handle)


nicolargo:
Ok, i just pull another commit (see: 781dfee) to add the 'key' in the dict because it is needed by Glances for the export process.

PR #1006, opened 2017-01-24
[WebUI] Hide public address if not available

notFloran:
#### Description

In the WebUI hide the public address if not available.

#### Resume

* Bug fix: yes
* New feature: no
* Fixed tickets: -


nicolargo:
Thanks @notFloran !

PR #1008, opened 2017-01-26
Fix typo in network text

HelioCampos:
form -> from

#### Description

Fix typo in network text. It was writen "form" when I think you meant "from".

#### Resume

* Bug fix: no
* New feature: no
* Fixed tickets: None


nicolargo:
Thanks for the head up !

PR #1081, opened 2017-04-16
Hide virtual devices in the diskio plugin

d10n:
#### Description

Hide virtual devices like dm-0 and loop0 by default in the diskio plugin, because:
* Disk IO from virtual / mapped devices is already accounted for in the real devices.
* Having many raw disks with many virtual devices makes the diskio plugin too tall

hide_virtual_block_devices

#### Resume

* Bug fix: no
* New feature: yes
* Fixed tickets:

### DIFF
--- a/docs/cmds.rst
+++ b/docs/cmds.rst
@@ -294,6 +294,10 @@ Command-Line Options
 
     show RAM FS in the DiskIO plugin
 
+.. option:: --diskio-show-virtual-devices
+
+    show virtual devices in the DiskIO plugin
+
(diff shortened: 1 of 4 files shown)


nicolargo:
Hi @d10n thnaks for the patch but you can do it directly via the Glances configuration file through the [diskio] section:

```
[diskio]
# Define the list of hidden disks (comma-separated regexp)
hide=dm-.*
```

d10n:
IMHO these should be hidden by default, but configuration is OK

PR #1107, opened 2017-06-13
Frame does not support the buffer interface

GoMapur:
The original code is causing "Frame does not support the buffer interface" error in zeroMQ @ https://github.com/zeromq/pyzmq/blob/master/zmq/sugar/socket.py line 373 in socket.recv_multipart. I looked up the code and find out the function b() in glances.compat is actually returning the same stuff. So it's basically a BUG of str does not support buffer interface, which can be solved by encode str before sending.

#### Description

#### Resume

* Bug fix: yes/no
* New feature: yes/no

### DIFF
--- a/glances/exports/glances_zeromq.py
+++ b/glances/exports/glances_zeromq.py
@@ -56,7 +56,7 @@ class Export(GlancesExport):
         self.client = self.init()
 
     def init(self):
-        """Init the connection to the CouchDB server."""
+        """Init the connection to the ZeroMQ server."""
         if not self.export_enable:
             return None
 
@@ -96,8 +96,8 @@ class Export(GlancesExport):
         # - First frame containing the following prefix (STRING)
         # - Second frame with the Glances plugin name (STRING)
         # - Third frame with the Glances plugin stats (JSON)
-        message = [b(self.prefix),
-                   b(name),
+        message = [b(self.prefix).decode('utf-8'),
+                   b(name).decode('utf-8'),
                    asbytes(json.dumps(data))]
 
         # Write data to the ZeroMQ bus


nicolargo:
Hi thanks for the patch, instead of using the .decode() function, i think that it is better to use the dedicated nativestr function available in the compat.py file (https://github.com/nicolargo/glances/blob/develop/glances/compat.py#L95). Compatible with both Python 2 and Python 3.

Just import:

    from glances.compat import range, nativestr

and use it.

GoMapur:
Hi thanks for the fast response. Sorry this is my very first few into open source community. I'll make the pull request again and ty for letting me know.

GoMapur:
I checked my pull request, I made the wrong change, the correct one should be 
```python
str.encode("utf-8")
```
Do you have any suggestion which method should I use in compat to encode the string?

nicolargo:
nativestr ?

Double check on Python 2 and Python 3. Encoding is a mess in Python...

Another think, please pull request on the DEVELOP branch, not the MASTER.

GoMapur:
Okay I'm on it. It is a very strange bug. The python version I'm using is 2.7.9. So the function b in compat.py is returning the same thing, but this results in the mentioned problem. However, after more investigation this, I found this should only happen in python3. I'll do more research, maybe this problem is not as simple as the solution of adding the encoding.

GoMapur:
Hi I find the issue. It turns out the string we have for ZeroMQ is unicode type, and you are right, to solve this for python 2 we only need to use nativestr. But this will be wrong in python3 since using nativestr in python3 will return a str back, which does not support buffer interface either.

PR #1108, opened 2017-06-13
Python encoding in python2 and python3

GoMapur:
In python2, binary type is native string type, so function b and nativestr should be the same. In python3, unicode is native string and binary is encoded unicode type. Now the question comes, why do we encode unicode string in python3 by encode('latin-1') instead using utf-8?

Note: I fixed the bug for ZeroMQ by modifying the compat file, still needs further test system wide to make sure the type checking and translating is totally correct.

#### Description

The original code is causing "Frame does not support the buffer interface" error in ZeroMQ @ https://github.com/zeromq/pyzmq/blob/master/zmq/sugar/socket.py line 373 in socket.recv_multipart. I looked up the code and find out the function b() in glances.compat is actually returning the unicode object. So it's basically a bug of unicode does not support buffer interface, which can be solved by encode str before sending.

#### Resume

* Bug fix: yes
* New feature: no

### DIFF
--- a/glances/compat.py
+++ b/glances/compat.py
@@ -85,7 +85,9 @@ if PY3:
         return iter(d.values())
 
     def u(s):
-        return s
+        if isinstance(s, text_type):
+            return s
+        return s.decode('utf-8', 'replace')
 
     def b(s):
         if isinstance(s, binary_type):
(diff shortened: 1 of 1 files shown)


nicolargo:
@GoMapur Lots of previous issues (https://github.com/nicolargo/glances/search?q=encoding&type=Issues&utf8=%E2%9C%93) came from encoding. 

Before merging your pull request, can you identify the string causing the initial issue ? 

I want to be sure that there is no side effect...

nicolargo:
@GoMapur Can you dump the content of data (or name or self.prefix) when the error is triggered ?

GoMapur:
@nicolargo when I debug I looked it up. It's not the content triggering the bug, it's b(self.prefix), which should be b("G") supposedly triggered the bug. Somehow my python2 is using unicode here instead of python2's native string, which should be binary type.

nicolargo:
Thanks for the PR !

PR #1114, opened 2017-06-23
fix min_size to not output string with more than max characters

vpoverennov:
#### Description
problem: min_size outputs names longer than max
old: min_size('some_name', 8) -> "_ome_name"
solution: add 1 to substring offset
new: min_size('some_name', 8) -> "_me_name"

#### Resume

* Bug fix: yes
* New feature: no
* Fixed tickets:


nicolargo:
Thanks !

PR #1158, opened 2017-10-01
 Prevent data truncation when receiving data from hddtemp server

mendelgusmao:
#### Description

~~It looks like hddtemp is a very non-deterministic piece of software and
does not grant that every connection will receive data for all devices
that were in the previous one, thus making a device and its temperature
disappear and reappear in the list once in a while.~~

Since the first implementation of the hddtemp plugin, Glances wasn't retrieving
all data from the hddtemp server, making the device disappear and reappear
in the list once in a while.

#### Resume

* Bug fix: yes
* New feature: no
* Fixed tickets: none


nicolargo:
Thanks for the contrib.

Merged to the DEVELOP branch.

PR #1172, opened 2017-10-23
Fixes a link rendering problem in load.rst

python33r:
Fixes a typo that prevented 'load on BSD' from rendering properly as a link.

#### Description

In the documentation, 'load on BSD' should render as a link but doesn't, due to a spurious `S` at the start - presumably a typo.

#### Resume

* Bug fix: yes
* New feature: no
* Fixed tickets: none


nicolargo:
Thanks for the patch !

PR #1186, opened 2017-11-14
Added using locale settings to encode wifi ssid. Replace any unknown …

### LINKED ISSUE #1185: 'ascii' codec can't encode character u'\U0001f4a9' in position 4: ordinal not in range(128)
With the wifi module enabled, I get an uncaught exception due to a neighbor's weird wifi name.     

The exact exception I get is:
`UnicodeEncodeError: 'ascii' codec can't encode character u'\U0001f4a9' in position 4: ordinal not in range(128)`     

The issue is caused from the following line of code (glances_wifi.py:200)
`msg = '{:{width}}'.format(hotspotname, width=ifname_max_width)`    

The cause being that the default encoding for strings when using format is 'ascii'. This won't be able to handle all characters that are allowed in a ssid (i.e. my neighbors wifi which is: Grey💩On) . 

The code could handle this by replacing unknown characters with '?' or by trying to use UTF-8 instead.

jkleve:
…characters in wifi ssid with '?'.

#### Description

Handle wifi ssid's with explicit string encoding or replace the character with '?' instead of crashing.

#### Resume

* Bug fix: yes
* New feature: no
* Fixed tickets: Fixes #1185

### DIFF
--- a/glances/plugins/glances_wifi.py
+++ b/glances/plugins/glances_wifi.py
@@ -19,12 +19,13 @@
 
 """Wifi plugin."""
 
+import locale
 import operator
+import psutil
 
 from glances.logger import logger
 from glances.plugins.glances_plugin import GlancesPlugin
 
-import psutil
 # Use the Wifi Python lib (https://pypi.python.org/pypi/wifi)
 # Linux-only
 try:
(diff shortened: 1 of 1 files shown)


nicolargo:
Thanks, corrected using the already existing u() function.

PR #1216, opened 2018-01-09
Don't load public IP address if module disabled

alyssais:
#### Description

glances shouldn't hit the network if there's no need to. This is especially annoying for me because I monitor outgoing network traffic and get an alert every time I run glances, even with `--disable-ip`.

#### Resume

* Bug fix: yes
* New feature: no


nicolargo:
Reviewed and merged in the develop branch (will be available in the future Glances 3.0 release).

Thanks for your contribution !

PR #1219, opened 2018-01-17
update readme docker run command

oppianmatt:
slightly better way which:
 - doesn't leave the container around when finished
 - uses network host mode which then allows glances to show the network devices

#### Description

improve docker run command in readme


nicolargo:
Thanks for the contribution !

nicolargo:
Also merged in the DEVELOP branch.

PR #1223, opened 2018-01-19
Dockerfile rebase to bitnami/minideb

dschaper:
#### Description

The docker image now weighs in at a decent 600M. Basing off of Bitnami's Minideb and getting a tarball image of Master from the GitHub api reduces the size to a light 150M. Might be able to reduce that even more by freezing the python into a bin and using just the 50M Minideb base. 

#### Resume

* Bug fix: no
* New feature: yes
* Fixed tickets: None

### DIFF
--- a/docker/master/Dockerfile
+++ b/docker/master/Dockerfile
@@ -7,10 +7,6 @@
 # Pull base image.
 FROM bitnami/minideb:stretch
 
-# Install Glances (develop branch)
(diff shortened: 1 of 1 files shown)


dschaper:
If this looks good I can squash that extra commit.

nicolargo:
Yes it is huge but the image include all the Python libs (and so all the Glances functions).
Perhaps it will be better to create a new "thin" image beside the stable and develop Dockerfile.

PR #1229, opened 2018-02-10
Implement number and alias aware diskio sorting

Thomaash:
#### Description

Sorts Disk I/O by aliases (if present) or device names and treats numbers as numbers.
Prevents situations such as `sda10` being above `sda2` or `alias` between `sda1` and `sda3`.

#### Resume

* Bug fix: no
* New feature: no
* Fixed tickets: none


nicolargo:
Many thanks for the PR !

Issue #1241, opened 2018-02-20
[WEB UI] Flood if refresh_time is negative


spike008t:
#### Description

On the web ui, the route `/<refresh_time:int>` can define the `refresh_time` on the webap.
But if we set a negative value (ie: `http://localhost:61208/-1`). We flood the server to refresh

<img width="1133" alt="screen shot 2018-02-20 at 23 10 44" src="https://user-images.githubusercontent.com/262459/36452192-d335e76c-1693-11e8-8719-3ea3afd311be.png">
<img width="559" alt="screen shot 2018-02-20 at 23 16 58" src="https://user-images.githubusercontent.com/262459/36452281-2e755824-1694-11e8-8a3a-33ca50a13d21.png">

The webapp should check the refresh delay and if that is < 1s, set to 1s or 5s.

#### Versions

* Glances (glances -V): latest develop version
* PSutil (glances -V):
* Operating System (lsb_release -a): 

#### Logs

You can also pastebin the Glances logs file (https://glances.readthedocs.io/en/latest/config.html#logging)

notFloran:
@spike008t thanks for the report, I have made a PR #1242

nicolargo:
@spike008t Can we close this issue ?

PR #1242, opened 2018-02-21
Fix WebUI when refresh_time is negative

notFloran:
#### Description

The WebUI now use the default refresh time if the value < 1.

#### Resume

* Bug fix: yes
* New feature: no
* Fixed tickets: #1241


nicolargo:
Thanks !

PR #1253, opened 2018-04-25
Use https URLs for checking external IP

viq:
Processing JSON from connection that could be changed by attacker in path sounds scary, and all those services serve over https as well.

#### Description

Check own IP over https. Glances processes output returned, which sometimes is JSON, and doing that over plaintext http which could be transparently modified in flight is scary.

#### Resume

* Bug fix: yes
* New feature: no


nicolargo:
Done in commit 8b6b0d3

PR #1323, opened 2018-09-23
Fix rounding of alert pourcent in the WebUI

notFloran:
##### Before

capture d ecran 2018-09-23 a 10 35 46

##### After

capture d ecran 2018-09-23 a 10 40 14

#### Resume

* Bug fix: yes
* New feature: no
* Fixed tickets: -


nicolargo:
Thanks @notFloran !

PR #1357, opened 2018-11-27
`pwd`.conf instead of ./glances.conf for docker run command

EmilienMottet:
#### Description

improve documentation of glances with docker.

My Docker version 18.09.0, build 4d60db4

if I run this 
```bash
docker run -v ./glances.conf:/glances/conf/glances.conf -v /var/run/docker.sock:/var/run/docker.sock:ro --pid host -it docker.io/nicolargo/glances
``` 

I have this error `docker: Error response from daemon: create ./glances.conf: "./glances.conf" includes invalid characters for a local volume name, only "[a-zA-Z0-9][a-zA-Z0-9_.-]" are allowed. If you intended to pass a host directory, use absolute path.`

So in agreement with docker document https://docs.docker.com/storage/bind-mounts/, we can replace ./ by \`pwd\`

#### Resume

\`pwd\`.conf instead of ./glances.conf for docker run command in documentation

* Bug fix: yes
* New feature: no

### DIFF
--- a/README.rst
+++ b/README.rst
@@ -192,9 +192,9 @@ docker run options:
 
 .. code-block:: console
 
-    docker run -v ./glances.conf:/glances/conf/glances.conf -v /var/run/docker.sock:/var/run/docker.sock:ro --pid host -it docker.io/nicolargo/glances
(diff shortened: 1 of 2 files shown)


nicolargo:
Thanks for the contribution !

PR #1390, opened 2019-01-11
README: note that docker plugin works on Mac

apjanke:
#### Description

The docker plugin sure seems to work on macOS, not just Linux. I tested it locally on macOS 10.13.6. Should the README be updated to note that it works there too?

#### Resume

* Bug fix: no
* New feature: no
* Fixed tickets: none


nicolargo:
Thanks !

Issue #1272, opened 2018-06-14
export to elasticsearch is broken for elastic 6.x


AllegroDvt:
[2018-06-14T15:46:42,117][DEBUG][o.e.a.a.i.m.p.TransportPutMappingAction] [zzz] failed to put mappings
 on indices [[[glances/QL4r1yDvRmi-shKTYkmq-A]]], type [sensors]                                         
java.lang.IllegalArgumentException: Rejecting mapping update to [glances] as the final mapping would have
 more than 1 type: [sensors, cpu]

That's related to https://www.elastic.co/guide/en/elasticsearch/reference/6.0/removal-of-types.html

nicolargo:
As a workaround can you try to deisable the sensor plufin (--disable-sensors)

genevera:
@nicolargo that doesn't make any difference.

genevera:
#1438 is a quick fix that forces all fields to go in as 'keyword' type.  If anyone knows a way to easily map Py3 types to ES types please lmk:)

edgeofthex:
> #1438 is a quick fix that forces all fields to go in as 'keyword' type. If anyone knows a way to easily map Py3 types to ES types please lmk:)

In exporting data from Glances 3.1.0 and 3.1.1b0 to  Elasticsearch 6.8 and 7.2, I encountered the same issues as @AllegroDvt regarding mapping types.  For me, the above quick fix didn't work.

I don't code, but am willing to contribute to a fix in other ways.  Lmk.

PR #1438, opened 2019-03-13
add date to index, unbreak object push

genevera:
#### Description

makes indices date-based (e.g. `glances-2019.03.13`) and also makes all fields of type `keyword`.  That is a hack and I will probably try to fix it later.

#### Resume

* Bug fix: yes
* New feature: no
* Fixed tickets: comma-separated list of tickets fixed by the PR, if any #1272


nicolargo:
Thanks for the contribution

genevera:
I realize i may have broken some dashboards by renaming the timestamp element.  I'll pr a change for that.

PR #1439, opened 2019-03-13
fix merge conflict in man page

oefd:
For whatever reason a merge conflict ended up in master - this just resolves that conflict.

This has filtered down at least to the Arch Linux package, but is also a purely aesthetic problem so whether it's worth publishing a new version and trying to get it repackaged isn't too clear cut.

I figure worth fixing it in master in the repo at least.

### DIFF
--- a/docs/man/glances.1
+++ b/docs/man/glances.1
@@ -1,10 +1,6 @@
 .\" Man page generated from reStructuredText.
 .
-<<<<<<< HEAD
 .TH "GLANCES" "1" "Jan 19, 2019" "3.1.0" "Glances"
-=======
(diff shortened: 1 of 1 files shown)


nicolargo:
Please PR should be done on DEVELOP branch

oefd:
Sorry if I was unclear, but this is an issue that was added to the master branch (and the master branch only) by b668b0bf4c54d082538f938f52f3093543fdd693 when a merge of develop into master accidentally left in some merge conflict markers.

All this PR is meant to do is hotfix it so you don't need to wait for the next merge of develop into master  to make the generated man page not render with some errant markers.

PR #1463, opened 2019-04-10
snap: Add packaging-specific gitignore rules

Lin-Buo-Ren:
#### Description
Avoid showing snap build artifacts as untracked files, this should also make the -dirty tag in the snap version string disappear.

#### Resume
* Bug fix: kinda
* New feature: no


nicolargo:
Thanks for the contribution !

PR #1471, opened 2019-04-19
Filter duplicate public IP addresses in IP plugin

mendelgusmao:
#### Description

I just noticed that httpbin.org is returning duplicate IP addresses in the format "1.2.3.4, 1.2.3.4". Here is a cosmetic fix that deduplicates the data. 

#### Resume

* Bug fix: yes
* New feature: no
* Fixed tickets: None


nicolargo:
Thanks for the enhancement @mendelgusmao

Issue #1420, opened 2019-02-11
[WEB-UI] Glances-3.1.0: no Linux distribution name in the web GUI


spezialist1:
#### Description

It looks like a bug: after the standard installation (from https://nicolargo.github.io/glances/) of the stable release Glances 3.1.0 with psutil 5.5.0 under CentOS-7.6, the Linux distribution name is not displayed in the web-GUI. Below is my screenshot where the top is the web-GUI, and the bottom is the usual ssh-console. It is clearly seen, that the Linux distribution name is present in the ssh-console:
![no_distro_name](https://user-images.githubusercontent.com/10425179/52559553-31165100-2dfe-11e9-9ea4-bcaabefddc10.png)

#### Versions

* Glances & psutil (glances -V): Glances v3.1.0 with psutil v5.5.0
* Operating System (lsb_release -a): CentOS Linux release 7.6.1810 (Core)

Very thanks for the support.

nicolargo:
@notFloran Can you have a look ? i think that the distrib name is available in the API.

PR #1472, opened 2019-04-20
[WebUI] Fix use of isLinux

notFloran:
#### Description

After the big refactoring of the webui, I broke the use of `isLinux`.

#### Resume

* Bug fix: yes
* New feature: no
* Fixed tickets: #1420


nicolargo:
Thanks !

Issue #1434, opened 2019-03-11
Content-security-policy blocking Glances web UI


leem32:
My Content-security-policy (CSP) is blocking the Glances web UI or more precisely it's preventing the inline script from loading because it does not contain a nonce.

```javascript
<script>
        angular
            .module('glances.config', [])
            .constant('REFRESH_TIME', 30);
</script>
```
Error:
```Uncaught Error: [$injector:modulerr] Failed to instantiate module glancesApp due to:
Error: [$injector:modulerr] Failed to instantiate module glances.config due to:
Error: [$injector:nomod] Module 'glances.config' is not available! You either misspelled the module name or forgot to load it. If registering a module ensure that you specify the dependencies as the second argument.
```

If I turn off the CSP Glances works without any issues.

I need to add a nonce into all inline scripts so the Glances inline script would look like this.
```javascript
<script nonce="<?php echo $nonce; ?>">
        angular
            .module('glances.config', [])
            .constant('REFRESH_TIME', 30);
</script>
```

Understandably you're not going to add the nonce into that script just for me, so I need to find another way around this problem.

I notice `glances.js` is not inline so that works OK with my CSP.

Can I load that script any other way so my CSP stops complaining?

nicolargo:
@notFloran Any advise concerning this request ?

notFloran:
I think #1473 fix the problem. @leem32 can you try the develop branch ?

leem32:
Yep, it's working with the develop branch. Thanks for fixing :)

PR #1473, opened 2019-04-20
Better method to pass "refresh time" from glances to the WebUI

notFloran:
#### Description

I rewrite some parts of the WebUI to pass the refreshTime variable from bottle to the WebUI without using an inline script in the HTML.

#### Resume

* Bug fix: no
* New feature: no
* Related tickets: #1434


nicolargo:
Nice !

Issue #1428, opened 2019-02-28
Glances crashes on docker restart


danst0:
I have glances open in one tmux session, in the other session I restart docker (service docker restart). Glances crashes.

nicolargo:
Is there a relevant message in the console or in the Glances log file ?
What is your Glances / Docker version.

Please fill the issue with all the information needed to investigate.

PR #1496, opened 2019-07-02
Improve error handling when docker restarts

joeabbey:
#### Description

Don't print errors to the console when docker restarts.  Instead, loss of connection to the daemon should stop the threads and clear out current version and container lists.

#### Resume

* Bug fix: yes
* New feature: no
* Fixed tickets: #1428 (guessing given the limited information in the issue)


nicolargo:
Thanks for the contrib !

PR #1507, opened 2019-07-14
Fix missing future in devel-arm Dockerfile

Chris-V:
#### Description

`future` has been added to requirements.txt, but is missing from arm Dockerfile.

Instead of adding future directly, I install all requirements so this should not be a problem in the future. Another improvement could be to install optional-requirements.txt so all modules are made available in the base image.

#### Resume

* Bug fix: yes
* New feature: no
* Fixed tickets: n/a


nicolargo:
Sorry for the delay. Hard work in the "real" life.

Thanks for the contrib !

Issue #1521, opened 2019-07-31
Increase web refresh rate


Paytah232:
From what I can see the web interface refreshes every 5 seconds.
Is there an environment variable I can set that will reduce that refresh rate to 1s or less?

a2902793:
`glances -t 1`

Paytah232:
I tried this and it seems to have increaseds the interval from 5secs to 6sec. Any other solution that could work? @nicolargo @a2902793

CanisW:
e.g. using http://0.0.0.0:61208/30 can set the refresh rate to every 30 seconds.
ref:
1. "Web Server Mode" section on https://glances.readthedocs.io/en/stable/quickstart.html
2. https://github.com/nicolargo/glances/issues/784

notFloran:
It seems that `glances -w -t 30` does not work anymore. I will look at it.

notFloran:
@Paytah232 can you try the 3.1.3 version ? is now possible to do this : `glances -w -t 1`

Paytah232:
How do I update on the synolgy nas? Is it auto or do I need to build it again from image?
________________________________

PR #1548, opened 2019-10-08
Fix custom refresh time in the web UI

notFloran:
#### Description

The cmd `glances -t 30 -w` was not working because in web UI mode, the time arg was force to 5 seconds.

#### Resume

* Bug fix: yes
* New feature: no
* Fixed tickets: #1521


nicolargo:
Perfect @notFloran !

Issue #1555, opened 2019-11-07
GPU stats not showing on mobile web view


itzsimpl:
I have the latest version of glances installed and running it in web server mode `glances -w`. If I access the web server from a desktop computer, the GPU stats are visible, but if I access it from a mobile (vertical or horizontal view) they are not displayed. Same is true if I access the web server from a pad in vertical view (stats are not displayed), however in a horizontal view they are.

nicolargo:
@notFloran Can you have a look on it ?

PR #1558, opened 2019-11-13
Display gpu and mem more plugin on small devices

notFloran:
#### Description

Display GPU and mem more plugins on small devices.

#### Resume

* Bug fix: yes
* New feature: no
* Fixed tickets: #1555


nicolargo:
Thanks @notFloran !

PR #1559, opened 2019-11-14
Bug fix for UTF-8 check

yoganandc:
#### Description

Pretty straight-forward, e was not defined causing glances to crash when encoding is not UTF-8

With this fix, a warning is appended to the log. Reproduced below:
```
2019-11-14 05:27:38,350 -- WARNING -- UTF-8 is mandatory for sparklines ('ascii' codec can't encode characters in position 0-39: ordinal not in range(128))
```

#### Resume

* Bug fix: yes
* New feature: no
* Fixed tickets: (none)


nicolargo:
Thanks @yoganandc !

Issue #1569, opened 2019-12-13
Unhandled AttributeError when no config files found


qubidt:
#### Description

**update**: description incorrect, see https://github.com/nicolargo/glances/issues/1569#issuecomment-565742311

---

Without any config files defined, [this logger call](https://github.com/nicolargo/glances/blob/e29a153d93a740397d2c7cfa256596240b9bc2b9/glances/plugins/glances_plugin.py#L89-L90) triggers an `AttributeError`.

It looks like this was introduced by a61c91b111f06523e8f265b7fcd63284833ad0dc.

I would write a PR but I'm not sure whether you'd like this resolved by fixing the `logger.debug` call or by providing a `Config` object when there are no config files to read. (I assume the former, though)

Stacktrace and logs included below.

#### Versions

* Glances & psutil (glances -V): `Glances v3.1.3 with psutil v5.6.7`
* Operating System (lsb_release -a): 
```
LSB Version:    1.4
Distributor ID: Arch
Description:    Arch Linux
Release:        rolling
Codename:       n/a
```

#### Logs

```
$ glances -c localhost
Traceback (most recent call last):
  File "/usr/bin/glances", line 11, in <module>
    load_entry_point('Glances==3.1.3', 'console_scripts', 'glances')()
  File "/usr/lib/python3.8/site-packages/glances/__init__.py", line 143, in main
    start(config=config, args=args)
  File "/usr/lib/python3.8/site-packages/glances/__init__.py", line 112, in start
    mode.serve_forever()
  File "/usr/lib/python3.8/site-packages/glances/client.py", line 238, in serve_forever
    if not self.login():
  File "/usr/lib/python3.8/site-packages/glances/client.py", line 162, in login
    if not self._login_glances():
  File "/usr/lib/python3.8/site-packages/glances/client.py", line 130, in _login_glances
    self.stats.set_plugins(json.loads(self.client.getAllPlugins()))
  File "/usr/lib/python3.8/site-packages/glances/stats_client.py", line 59, in set_plugins
    self._plugins[item] = plugin.Plugin(args=self.args)
  File "/usr/lib/python3.8/site-packages/glances/plugins/glances_connections.py", line 60, in __init__
    super(Plugin, self).__init__(args=args,
  File "/usr/lib/python3.8/site-packages/glances/plugins/glances_plugin.py", line 90, in __init__
    config.config_file_paths()))
AttributeError: 'NoneType' object has no attribute 'config_file_paths'
```

```
[...]

PR #1571, opened 2019-12-14
Fix broken method call when config is None

qubidt:
#### Description

Fix method calls on `config` triggering an `AttributeError` when no config files were found

#### Resume

* Bug fix: yes
* New feature: no
* Fixed tickets: #1569


nicolargo:
Merged.

Thanks for the PR !

PR #1596, opened 2020-02-07
Add hide option to sensors plugin

cdeletre:
#### Description

This pull request proposes to add the support of hide option to the sensors plugin. It may be useful when you want to hide core temperatures and only show main CPU temp of 10+ core CPUs.

#### Resume

* Bug fix: no
* New feature: yes
* Fixed tickets: N/A

### DIFF
--- a/glances/plugins/glances_sensors.py
+++ b/glances/plugins/glances_sensors.py
@@ -127,14 +127,20 @@ class Plugin(GlancesPlugin):
 
             pass
 
-        # Set the alias for each stat
+        # Update the stats
+        self.stats = []
+
         for stat in stats:
+            # Do not take hide stat into account
+            if self.is_hide(stat["label"]):
+                continue
+
+            # Set the alias for each stat
             alias = self.has_alias(stat["label"].lower())
             if alias:
                 stat["label"] = alias
 
-        # Update the stats
-        self.stats = stats
+            self.stats.append(stat)
 
         return self.stats
 


glances/plugins/glances_sensors.py
```diff
@@ -127,14 +127,20 @@ def update(self):
 
             pass
 
-        # Set the alias for each stat
+        # Update the stats
+        self.stats = []
```
cdeletre: I had no idea what would be the best, and good, way to implement the hide option. I have supposed there is no sync issue doing this way, using no temporary variable.

nicolargo:
Done in version 3.1.4.

Thanks for the PR @cdeletre

modelrockettier:
@nicolargo is it possible to reapply this pull?

It looks like somehow it didn't make it into either version 3.1.4 or the develop branch.

distributorofpain:
Does this mean i can set the alias=hide and it will hide the senor?  Assuming its put back in 3.1.5...

modelrockettier:
Not exactly, but similar.  It follows the other plugin style of:
```
hide=Sensor 1,Sensor 3
```

distributorofpain:
oh, thanks for the clarification!  That makes perfect sense, i hadnt used the hide function yet and when i searched the docs for hide sensor it wasnt shown as available(so i didnt try it).

I can confirm its working in:  Glances v3.1.5 with PsUtil v5.7.2  ( i just installed it yesterday on this machine )

PR #1598, opened 2020-02-09
glances can work in console mode on windows - windows-curses

bartlomiejcieszkowski:
#### Description

We can have glances working on windows in console mode - there is windows-curses module which works pretty great, snip from w10 with pip installed windows-curses:

windows-curses_glances

### DIFF
--- a/glances/main.py
+++ b/glances/main.py
@@ -305,10 +305,6 @@ Examples of use:
         if args.disable_autodiscover:
             logger.info("Auto discover mode is disabled")
 
-        # By default Windows is started in Web mode
-        if WINDOWS:
-            args.webserver = True
-
         # In web server mode
         if args.webserver:
             args.process_short_name = True
--- a/glances/outputs/glances_curses.py
+++ b/glances/outputs/glances_curses.py
@@ -31,14 +31,15 @@ from glances.processes import glances_processes
 from glances.timer import Timer
 
 # Import curses library for "normal" operating system
-if not WINDOWS:
-    try:
-        import curses
-        import curses.panel
-        from curses.textpad import Textbox
-    except ImportError:
-        logger.critical("Curses module not found. Glances cannot start in standalone mode.")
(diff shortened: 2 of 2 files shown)


nicolargo:
Last time i try it, the other plugins (processes list, sensor, disk...) are not displayed as expected. Please can you test it ?

bartlomiejcieszkowski:
Sure,
i can see the of by 1 size issue - did locally a w/a which adjusted width by -1 and it worked okay, but thats not a way to go.

Doing some debugging and checking the window-curses and pdcurses underneath it for why the size is off

bartlomiejcieszkowski:
ive checked the screen sizes i get on windows
screen_x resolves to 131

which is bad, as the window has only 130 usable cols, 
but the created term_window has a correct size on both windows and ubuntu, and looks like a better place to get available screen size as we are working on it

so i changed to use the term_window as source of getmaxyx

bartlomiejcieszkowski:
image

^ end result
only issue is that the windows prints long name of timezone hence weird spacing, but thats the different issue :)

bartlomiejcieszkowski:
the long timezone is returned on windows for timezone.tzname, from what i checked under the hood python uses strftime("%Z") (https://github.com/python/cpython/blob/master/Modules/timemodule.c#L1538)  which returns long string on windows, which is kinda ok since %Z produces:

> Timezone name or abbreviation *If timezone cannot be determined, no characters

i could try the %z which would result in +/-HHMM

i.e:
```
>>> time.strftime('%z')
'+0100'
```

nicolargo:
Hi @bartlomiejcieszkowski 

Sorry for the delay...

I just merge your PR in the DEVELOP branch. 

I also patch the time zone by removing it on Windows OS.

Can you test it (no Windows computer on my side...).

bartlomiejcieszkowski:
@nicolargo tested, works :)

PR #1599, opened 2020-02-10
Specify precision to seconds for influxDB output 

fcalvet:
#### Description

Per the [documentation](https://docs.influxdata.com/influxdb/v1.7/write_protocols/line_protocol_reference/#syntax-description): 

> To significantly improve compression, use the coarsest precision possible for timestamps.

As we do not set it currently it defaults to nanoseconds. The library we use [supports](https://influxdb-python.readthedocs.io/en/latest/api-documentation.html#influxdb.InfluxDBClient.write_points) setting it in the write request.

This doesn't seem like opposing glances ideas as it seems built around seconds (eg you can set -t 0.1 but if your --export csv the timestamps are truncated to seconds).

**This pull request also includes:**
* Fixes GPU memory consumption reporting which was wrong after addition of temperature reporting:
```
>>> print(py3nvml.nvmlDeviceGetUtilizationRates(handle))
c_nvmlUtilization_t(gpu: 96 %, memory: 37 %)
>>> print(py3nvml.nvmlDeviceGetMemoryInfo(handle))
c_nvmlMemory_t(total: 11554717696 B, free: 2247688192 B, used: 9307029504 B)
```
See man page of nvidia-smi
```
   Utilization
       Utilization rates report how busy each GPU is over time, and can be used to determine how much an application is using the GPUs in the system.

       Memory         Percent of time over the past sample period during which global (device) memory was being read or written.  The sample period may be between 1 second and 1/6 second depending on the product.
```

* Small update to grafana dashboard to correctly calculate diskio (dependent on time_since_update)
* Small update to glances/README.txt, updated link to docs and added AMPs in file structure
* Document need of elasticsearch module for elasticsearch export (missing also for influxDB but there seems to be some kind of link collision due to rst)

PS: There's a bit of everything in this PR, if you prefer I can make separate ones, but I just went for fixing what I saw.

#### Resume

* Bug fix: yes
* New feature: no


nicolargo:
Thanks you for the PR !

PR #1655, opened 2020-05-07
Unittests: Use sys.executable instead of hardcoding the python interpreter

mnhauke:
#### Description

Use sys.executable instead of hardcoding the python interpreter in the unittests.

#### Resume

* Bug fix: yes
* New feature: no
* Fixed tickets: -


nicolargo:
Thanks for the PR @mnhauke !

Issue #1694, opened 2020-07-21
Enhancement: RSS for containers


anjz:
Hello,

I think it'd be nice to get the `rss` and `cache` stats for containers back. They were removed long time ago because apparently Docker 1.11 removed them. AFAIK this is no longer the case.

I made some changes to put them back and also to display an RSS column in the web UI for container stats, since I believe this is a useful metric to show.

Thanks

nicolargo:
Done in https://github.com/nicolargo/glances/pull/1695

Issue #848, opened 2016-04-14
Crash with Docker 1.11


dbarboza:
#### Description
Glances CLI errors with the following trace:

```
$ glances
Traceback (most recent call last):
  File "/usr/local/bin/glances", line 9, in <module>
    load_entry_point('Glances==2.6.1', 'console_scripts', 'glances')()
  File "/usr/local/lib/python3.4/dist-packages/glances/__init__.py", line 133, in main
    standalone.serve_forever()
  File "/usr/local/lib/python3.4/dist-packages/glances/standalone.py", line 107, in serve_forever
    return self.__serve_forever()
  File "/usr/local/lib/python3.4/dist-packages/glances/standalone.py", line 88, in __serve_forever
    self.stats.update()
  File "/usr/local/lib/python3.4/dist-packages/glances/stats.py", line 154, in update
    self._plugins[p].update()
  File "/usr/local/lib/python3.4/dist-packages/glances/plugins/glances_plugin.py", line 658, in wrapper
    ret = fct(*args, **kw)
  File "/usr/local/lib/python3.4/dist-packages/glances/plugins/glances_docker.py", line 226, in update
    container['memory'] = self.get_docker_memory(container['Id'], self.thread_list[container['Id']].stats)
  File "/usr/local/lib/python3.4/dist-packages/glances/plugins/glances_docker.py", line 297, in get_docker_memory
    ret['rss'] = all_stats['memory_stats']['stats']['rss']
TypeError: 'NoneType' object is not subscriptable
```

In the brief time glances is displayed, I noticed the docker stats are mostly "?" instead of numeric:
```
docker0       0b     0b    Name                               Status  CPU%    MEM IOR/s IOW/s  Rx/s  Tx/s Command 
eth0          0b     0b    gemstash_server          Up About an hour   0.0      ?     ?     ?     ?     ? gemstash start --no-daemonize
```

I have one container running and `docker stats gemstash_server` appears normal:
```
CONTAINER           CPU %               MEM USAGE / LIMIT     MEM %               NET I/O             BLOCK I/O           PIDS
gemstash_server     0.01%               35.75 MB / 25.23 GB   0.14%               4.994 kB / 648 B    0 B / 0 B           0
```
If I have no containers running, glances will run fine without the docker section.
I did upgrade docker to the newest this morning (v 1.11.0). I don't see anything obvious in the release notes that would have broken stats API: https://github.com/docker/docker/releases
[...]

PR #1695, opened 2020-07-21
Add RSS metric for containers

anjz:
#### Description

Add `rss` and `cache` memory stats for containers.

Docker plugin now returns `rss` and `cache` stats for containers. These were commented out due to them not being present in Docker 1.11 (maybe due to a bug in Docker?) See https://github.com/nicolargo/glances/issues/848. Since these stats are available in newer versions of Docker I thought it'd be nice to bring them back.

I've also added an additional column for `rss` in the web UI for containers.

#### Resume

* Bug fix: no
* New feature: yes
* Fixed tickets: Enhacement https://github.com/nicolargo/glances/issues/1694


nicolargo:
Thanks for the PR @anjz 

Merged. Will be available in the future releases.

PR #1722, opened 2020-09-09
Fix container rss value

anjz:
#### Description

Container RSS metric in the web UI was getting the value from container memory instead of container rss.

#### Resume

* Bug fix: yes
* New feature: no


notFloran:
You need to commit/push the builded assets, see https://github.com/nicolargo/glances/blob/develop/glances/outputs/static/README.md

anjz:
I've pushed the built assets, but I didn't have to do it when I first introduced the rss metric: https://github.com/nicolargo/glances/pull/1695

notFloran:
Thanks @anjz

Issue #1697, opened 2020-07-24
GPU information is incomplete via web


nick-stocks:
#### Description

GPU information is incomplete when viewed via the web portal, despite the API returning all the data. It works fine when running it via terminal. It is the same issue regardless of whether glances is running as my user or as a service.

Config located in /etc/glances

Via web:
![image](https://user-images.githubusercontent.com/22432139/88379165-53095680-cd9a-11ea-84b0-dc774d16e0d0.png)

When accessed via http://0.0.0.0:61208/api/3/gpu
`[{"key": "gpu_id", "gpu_id": 0, "name": "GeForce GTX 1050 with Max-Q Design", "mem": 2.5310529929467576, "proc": 0, "temperature": 42}]`

Via terminal
![image](https://user-images.githubusercontent.com/22432139/88379189-61f00900-cd9a-11ea-9035-49e7be46dafc.png)

#### Versions

* Glances v3.1.4.1 with PsUtil v5.7.2
* Operating System
Distributor ID:	Pop
Description:	Pop!_OS 20.04 LTS
Release:	20.04
Codename:	focal

#### Logs

https://pastebin.com/qSd4HmEM

nicolargo:
Hi @nick-stocks 

when i have a look on your log file, i saw that the GPU plugin update the stat to the following value :

```
2020-07-24 10:54:19,553 -- DEBUG -- Plugin gpu update return [{'key': 'gpu_id', 'gpu_id': 0, 'name': 'GeForce GTX 1050 with Max-Q Design', 'mem': 2.5310529929467576, 'proc': 0, 'temperature': 43}] in 0.000403 seconds
```
So the API return this value.

Can you show me a case, in the log file, where proc stats not equal 0 and the API return 0 ?

RobOo-Github:
I have a similar situation.
On the console, the GPU info looks fine:
![Screenshot_2020-09-01_11-23-41](https://user-images.githubusercontent.com/65445647/91832468-ba41e280-ec45-11ea-81db-10c5bd753fa4.png)
The api view on http://127.0.0.1:61208/api/3/gpu is also OK most of the time:
![Screenshot_2020-09-01_11-25-37](https://user-images.githubusercontent.com/65445647/91832618-eeb59e80-ec45-11ea-8733-83575e032db2.png)
but not always-this is where I see N/A after 'Proc' in the web interface:
[...]

PR #1773, opened 2020-12-20
[WebUI] Fix display of null values for GPU plugin

notFloran:
#### Description

The condition to display "N/A" was wrong and N/A was displayed when the value = 0.

#### Resume

* Bug fix: yes
* New feature: no
* Fixed tickets: #1697


nicolargo:
Thanks @notFloran !

PR #1785, opened 2021-01-04
Add option to set the strftime format

meganomic:
Add an option in the .conf and commandline to change the strftime format.

Set `ConfigParser(interpolation=None)` because otherwise you have to write `%%Y%%M` and so on. I don't think interpolation is used anywhere?

Also set the `[global]` config option `check_updates` to false by default. Since `[global]` exists, because `strftime_format` is set under it, **outdated.py** complains about `check_update` not being set otherwise.

Thoughts?


nicolargo:
Thanks for the PR @meganomic !

PR #1818, opened 2021-02-27
Update README.rst

stesser:
Documentation: Change package install command for FreeBSD

Python-2.7 is EOL and no longer supported in FreeBSD - the new default version is Python-3.7 and the package name has to be adjusted to refer to that Python version.


nicolargo:
Thanks for the PR !

Issue #1858, opened 2021-05-13
Docker image with additional  dependencies included?


Poeschl:
**Is your feature request related to a problem? Please describe.**
I came to this idea while struggling with missing python dependencies for the exporter on the latest docker image. I found the issues which suggest to use the dev image since it has the dependencies, but that one is also not versioned and the name suggests that is might not be as stable as a versioned release.

**Describe the solution you'd like**
Would it be possible to release also a docker images like `nicolargo/glances:3.1.7-full` where all the additional dependencies are included? That way there would be versioned release with all dependencies. It would be bigger then the regular one, but as user of it you have the choice between the two.

**Describe alternatives you've considered**
As alternative the current docker image can be used as base image and the python packages can be installed individually. This is possible, but it also need to be active maintained on new releases.

nicolargo:
It is already availble through the dev tag.

Latest(master with minimal plugins):
```
$ docker run --rm -v /var/run/docker.sock:/var/run/docker.sock:ro --pid host --network host -it nicolargo/glances:latest pip freeze
bernhard==0.2.6
bottle==0.12.19
certifi==2020.12.5
chardet==4.0.0
docker==5.0.0
future==0.18.2
Glances==3.1.6.2
idna==2.10
ifaddr==0.1.7
netifaces==0.10.9
pbkdf2==1.3
ply==3.11
protobuf==3.15.8
psutil==5.8.0
py-cpuinfo==4.0.0
py3nvml==0.2.6
pyasn1==0.4.8
pycryptodomex==3.10.1
pygal==2.4.0
pymdstat==0.4.2
pySMART.smartx==0.3.10
pysmi==0.3.4
pysnmp==4.4.12
pystache==0.5.4
requests==2.25.1
scandir==1.10.0
six==1.15.0
sparklines==0.4.2
urllib3==1.26.4
websocket-client==0.58.0
wifi==0.3.8
xmltodict==0.12.0
zeroconf==0.29.0
```

Dev with all plugins:
```
$ docker run --rm -v /var/run/docker.sock:/var/run/docker.sock:ro --pid host --network host -it nicolargo/glances:dev pip freeze
bernhard==0.2.6
bottle==0.12.19
cassandra-driver==3.24.0
certifi==2020.12.5
chardet==4.0.0
click==7.1.2
CouchDB==1.2
docker==4.4.1
elasticsearch==7.10.1
future==0.18.2
geomet==0.2.1.post1
idna==2.10
[...]

PR #1859, opened 2021-05-15
Docker images overhaul

Poeschl:
#### Description

The primary goal is the automatic publishing of the docker images in alpine/debian and minimal/full variants.
Achieving this I also made some improvements on the docker builds.

The docker builds are now multi-staged, that means the building of the dependencies is independent of the final image.
With that, all dependencies needed for the building are not included in the `glances` images.
For the `latest` images, this saves around 138MB compared with before.
Also the different stages of the Dockerfile can be build in parallel which will also increase build speed.

There is now the `*-full` docker image tag with all dependencies included for the exporting functions.
Also the images of glances have now a full alpine alternative, which is much smaller is size.

I also rewrote the GitHub Action für building to adjust to the new Dockerfiles.
Also the image building is done as an matrix build now, which also builds every image in parallel.

#### Resume

A Push to master, develop or a version branch will now trigger a docker images build after the python package is published on pypi. The images which are (as example a master-build): `glances:latest`, `glances:latest-full`, `glances:alpine-latest`, `glances:alpine-latest-full`.
The `glances:dev` image has no `-full` tag, since it includes already everything.

* Bug fix: no
* New feature: no
* Fixed tickets: #1858


docker-files/alpine.Dockerfile
```diff
@@ -0,0 +1,78 @@
+#
+# Glances Dockerfile (based on Alpine)
+#
+# https://github.com/nicolargo/glances
+#
+
+FROM alpine:3.13 as build
```
nicolargo, line 7: Do you think that it is necessary to fix the Alpine version inthe DockerFile ?

nicolargo (commented):
Good job ! Just a (stupid) question regarding the Alpine version.

docker-files/alpine.Dockerfile
Poeschl: Maybe don't exactly the alpine version, but somehow I wanted to stick with a known python version for the package copies between the stages in the Dockerfiles.
Without it the packages won't get copied on a python minor upgrade, since the folder structure is changing.

nicolargo:
Make sense !

Thanks for this great work @Poeschl

Poeschl:
You're welcome. 😉 Happy to help

nicolargo:
Hi @Poeschl 

There is an error while the build.yml is executed by Github action:

The error is in the "Build and push image" task:

```
...
#49 [auth] library/dev:pull,push token for registry-1.docker.io
#49 sha256:10284b64ef181163ee28a26717bf6ce6910fa79e16d4eeaa2821fa12fd0544b3
#49 DONE 0.0s

#50 [auth] library/dev:pull,push token for registry-1.docker.io
#50 sha256:e3fef1be84bc76ba3a95b478ccbf3bef1e63c165fbb1fd1ec407843ca3518222
#50 DONE 0.0s

#45 exporting to image
#45 sha256:e8c613e07b0b7ff33893b694f7759a10d42e180f2b4dc349fb57dc6b71dcab00
#45 pushing layers 0.4s done
#45 ERROR: server message: insufficient_scope: authorization failed
------
 > exporting to image:
------
error: failed to solve: rpc error: code = Unknown desc = server message: insufficient_scope: authorization failed
Error: buildx call failed with: error: failed to solve: rpc error: code = Unknown desc = server message: insufficient_scope: authorization failed
```

May it came from the old buidx job name rename create_Docker_builds ?

Can you have a look ?

Poeschl:
Yes, just took a quick look at the last build logs and it seems the `nicolargo/` prefix of the docker images is missing.
I will create a fixing MR tomorrow.

PR #1860, opened 2021-05-15
Fix missing docker image name

Poeschl:
#### Description

Fixes the failing docker build.

I forgot the `nicolargo/glances` prefix in my previous MR, so docker tried to push to `dev`.
Which is obviously not allowed.

#### Resume

Now the images have the image name `nicolargo/glances`.

* Bug fix: yes
* New feature: no


nicolargo:
Look great :)

image

Poeschl:
Looked at the CI log like it was a TV show 😂.
The build time should also decrease a lot the next time it gets built.

PR #1873, opened 2021-05-31
Fix history_add, so it allows history_max_size to be zero

metayan:
#### Description
Prevents pop from an empty list when history_max_size is zero.
The error was introduced in bd6e682a .

#### Resume
* Bug fix

#### Error fixed
Setting
```
history_size=0
```
in `glances.conf` caused
```
  File "/Users/Shared/src/scm/os/glances/glances/attribute.py", line 85, in value
    self.history_add(self._value)
  File "/Users/Shared/src/scm/os/glances/glances/attribute.py", line 110, in history_add
    self._history.pop(0)
IndexError: pop from empty list
```

### DIFF
--- a/glances/attribute.py
+++ b/glances/attribute.py
@@ -21,7 +21,6 @@
 
 from datetime import datetime
 
-
 class GlancesAttribute(object):
 
     def __init__(self, name, description='', history_max_size=None):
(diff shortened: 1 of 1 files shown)


nicolargo:
Thanks for the PR @metayan  !

PR #1894, opened 2021-07-13
Allow using maximum current frequency instead of average current frequency

goshawk22:
#### Description

Allow users to use the maximum current frequency of their CPU, as opposed to an average of all current CPU freqencies

Some users may find it more useful to see what frequencies their CPU is clocking up to. For example, some CPU's turbo much higher on a single core when running single core workloads.

#### Resume

* Bug fix: no
* New feature: yes
* Fixed tickets: None

### DIFF
--- a/conf/glances.conf
+++ b/conf/glances.conf
@@ -44,6 +44,9 @@ swap_careful=50
 swap_warning=70
 swap_critical=90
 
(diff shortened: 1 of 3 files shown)


nicolargo:
Thanks for the contribution @goshawk22 !

goshawk22:
@nicolargo was this ever merged?

PR #1954, opened 2021-11-18
Fix Grafana CPU temperature panel

NefixEstrada:
#### Description

It had a specific host, with this fix it will work for all hosts

#### Resume

* Bug fix: yes
* New feature: no


nicolargo:
Thanks for the patch @NefixEstrada  !

PR #1962, opened 2021-11-28
Typo fixes & Standardize doc string convention

RazCrimson:
#### Description

Clean up for doc-strings that don't match the convention specified and typo/spelling/grammar fixes wherever applicable.

An attempt at closing #940 

#### Resume

* Bug fix: No
* New feature: No

PR #1963, opened 2021-11-28
Typo fixes & Standardize doc string convention (contd.)

### LINKED ISSUE #940: Docstring convention
#### Description

This enhancement aims at uniformizing the docstings.

Docstrings
- Docstrings should not start with a space.
- Multi line docstrings should end on a new line.
- Multi line docstrings should start without a leading new line.
- Multi line docstrings should start with a one line summary followed by an empty line.

Example:

```
"""A multi line docstring has a one-line summary, less than 80 characters.

Then a new paragraph after a newline that explains in more detail any
general information about the function, class or method. Example usages
are also great to have here if it is a complex class or function.

When writing the docstring for a class, an extra line should be placed
after the closing quotations. For more in-depth explanations for these
decisions see http://www.python.org/dev/peps/pep-0257/

If you are going to describe parameters and return values, use Sphinx, the
appropriate syntax is as follows.

:param foo: the foo parameter
:param bar: the bar parameter
:returns: return_type -- description of the return value
:returns: description of the return value
:raises: AttributeError, KeyError
"""


```

RazCrimson:
#### Description

Continuation of #1962

#### Resume

* Bug fix: No
* New feature: No
* Closes #940

### DIFF
--- a/glances/__init__.py
+++ b/glances/__init__.py
@@ -78,7 +78,7 @@ def end():
         # ...after starting the server mode (issue #1175)
         pass
 
-    logger.info("Glances stopped (keypressed: CTRL-C)")
+    logger.info("Glances stopped (key pressed: CTRL-C)")
 
     # The end...
     sys.exit(0)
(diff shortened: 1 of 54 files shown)


RazCrimson:
That should close #940 

Let me know in case something needs a change :)

nicolargo:
Review and merfe on develop.

Thanks @RazCrimson !

PR #2029, opened 2022-04-10
Update public ip at intervals

### LINKED ISSUE #2028: Update external IP address
At the top of the display where:

     **machine.name (openSUSE Tumbleweed 20220408 64bit / Linux 5.17.1-1-default) - IP 192.168.0.150/24 Pub 111.111.111.111**

is shown the Pub 111.111.111.111 is only refreshed when glances is loaded. If that IP where to change in the case of a VPN address the new address is never refreshed.

I was wondering if this might be something that could be looked into?

RazCrimson:
#### Description

Let the Public IP address get updated every `t` seconds

The time interval can be configured by the config option under `ip` section with `public_refresh_interval`.
Default value is 300 seconds.

#### Resume

* Bug fix: no
* New feature: no
* Fixed tickets: Closes #2028

### DIFF
--- a/docs/aoa/header.rst
+++ b/docs/aoa/header.rst
@@ -11,6 +11,17 @@ Additionally, on GNU/Linux, it also shows the kernel version.
 
 In client mode, the server connection status is also displayed.
 
+It is possible to define time interval to be used for refreshing the
+public IP address (default is 300 seconds) from the configuration
+file under the ``[ip]`` section:
+
+.. code-block:: ini
+    [ip]
+    public_refresh_interval=240
+
+**NOTE:** Setting low values will result in frequent HTTP request to
+the IP detection servers. Recommended range: 120-600 seconds
+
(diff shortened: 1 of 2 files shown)


nicolargo:
Thanks for the PR @RazCrimson !

vvstubbs:
Okay this is really dumb. I cleaned out all my old installs so to do a fresh install from the recent git pull from the repository and installed all is good redid the glances.conf file and the IP information is not displayed, I press the [ I ] key and nothing. In the conf file 
[ip]
disable=False
remind and old confused brain

vvstubbs:
late breaking extra: But if I install as root IP works. as root of course installs in  /usr/bin/glances and uses the system python --version 3.8 and my user of course is local with my python --version 3.9, not a big deal but still odd to me

vvstubbs:
I works perfectly

PR #2082, opened 2022-07-09
fix: fix unit testing on macOS

jiegec:
#### Description

This pull requests fixes the unit tests on macOS.

In test_017_programs, the following exception may occur:

TypeError: unsupported operand type(s) for +=: 'int' and 'NoneType'

This commit fixes this exception by giving a default value for None.

#### Resume

* Bug fix: yes
* New feature: no
* Fixed tickets: none


nicolargo:
Thanks for the PR @jiegec !

PR #2088, opened 2022-07-14
Add note about `host` for Prometheus exporter in a container

### LINKED ISSUE #2087: Add example for using Prometheus Exporter functionality in a container
The [Prometheus Exporter documentation](https://glances.readthedocs.io/en/latest/gw/prometheus.html) `glances.conf` example is for when Glances is installed on the host, and doesn't work correctly when used in a container.

A note added to the docs with working configuration for usage in a container. To work in a container, the `host` needs to be set to `0.0.0.0` otherwise, you'll see errors in Prometheus similar to `connect: connection refused`.

I can raise a PR for this if that would help 🙂

dbrennand:
#### Description

Add a note to the Prometheus Exporter documentation about using `host=0.0.0.0` in the `glances.conf` file otherwise, it will not work.

#### Resume

* Bug fix: no
* New feature: no
* Resolves #2087

### DIFF
--- a/docs/gw/prometheus.rst
+++ b/docs/gw/prometheus.rst
@@ -15,7 +15,13 @@ listening on <host:port> (define in the Glances configuration file).
     prefix=glances
     labels=src:glances
 
-Note: you can use dynamc fields for the label (ex: labels=system:`uname -s`)
+.. note::
+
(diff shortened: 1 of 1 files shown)


nicolargo:
Thanks for the tips @dbrennand !

PR #2239, opened 2023-01-06
Consider quotes when splitting command arguments

mfridge:
#### Description

I want to execute shell commands in an action. Those commands are currently split by space character for security reasons (to me it's not clear why this required, but there are probably good reasons to do that ;)). As I would like to pass strings of multipel words to my commands, we have to consider quotes in command arguments.

One simple use case is the following action:
```
send_mail.sh "this is my subject" "and this is my alert text"
```

#### Resume

* Bug fix: yes
* New feature: no
* Fixed tickets:

### DIFF
--- a/glances/secure.py
+++ b/glances/secure.py
@@ -11,6 +11,7 @@
 
 from glances.compat import nativestr
 from subprocess import Popen, PIPE
+import re
 
 
 def secure_popen(cmd):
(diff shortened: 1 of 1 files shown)


nicolargo:
Thanks for the PR @mfridge !
Merge in the develop branch.

It will be release in Glances 3.4.0.

PR #2245, opened 2023-01-14
setup: include ujson as required dependency

chenrui333:
also update psutil version constraint

relates to https://github.com/Homebrew/homebrew-core/pull/120603

#### Description

Update install requires so that dependency updates can be inferred on the homebrew side (see linked PR)


chenrui333:
np.

PR #2255, opened 2023-01-22
moved now plugin

### LINKED ISSUE #2157: Can't see current time b/c it's beneath the processes list
Not to be a pain, but in web-server mode, when showing the current processes, the time is displayed at [the very bottom of the list](https://imgur.com/a/IWFzQnB) (as are the log messages). Previously, log messages were show above the list, and it makes it kinda unreadable having to scroll all the way down just to see the time

Is there anyway to change this? Thanks

I have a 16:9 1080p monitor if resolution affects layout at all

_Originally posted by @Cornelius-Figgle in https://github.com/nicolargo/glances/issues/2150#issuecomment-1282031514_

fr4nc0is:
fix #2157


nicolargo:
Thanks for the PR @fr4nc0is !

PR #2299, opened 2023-03-19
Make filesystem list unique

rhoriguchi:
#### Description

Remove duplicate entries in the `FILE SYS` list. For some reason ZFS datasets are shown more than once, this removes the duplicate entries and filters them by `mnt_point`.

Before:
```console
FILE SYS                 Used  Total
/                        351G  1.79T
/boot (nvme0n1p1)       20.5M   511M
/mnt/Data/Backup         375G  4.73T
/mnt/Data/Backup         375G  4.73T
/mnt/Data/Deluge        2.25M  4.36T
/mnt/Data/Deluge        2.25M  4.36T
/mnt/Data/Snapshots      256K  4.36T
/mnt/Data/Snapshots      256K  4.36T
_iners/storage/overlay   351G  1.79T
```

After:
```console
FILE SYS                 Used  Total
/                        351G  1.79T
/boot (nvme0n1p1)       20.5M   511M
/mnt/Data/Backup         375G  4.73T
/mnt/Data/Deluge        2.25M  4.36T
/mnt/Data/Snapshots      256K  4.36T
_iners/storage/overlay   351G  1.79T
```

#### Resume

* Bug fix: yes
* New feature: no
* Fixed tickets:

PR #2302, opened 2023-03-21
Plugin fs - cleaner impl to avoid duplicate entries when allow is set

RazCrimson:
Already fixed in 09c66999
A cleaner implementation of the fix..
In this case, I think fixing the cause of the bug is better than adding a cleanup at the end.

@nicolargo 
What do you suggest when looking at PRs from contributers (like #2299), 
Should I request for cleaner changes in cases where their fix works but could think of a better alternative/solution?
Or should I accept the changes and make the fix cleaner later when possible?  

What are your thoughts on this?

### DIFF
--- a/glances/plugins/glances_fs.py
+++ b/glances/plugins/glances_fs.py
@@ -13,6 +13,7 @@ from __future__ import unicode_literals
 import operator
 
 from glances.compat import u, nativestr, PermissionError
+from glances.logger import logger
 from glances.plugins.glances_plugin import GlancesPlugin
 
 import psutil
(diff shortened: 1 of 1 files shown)


nicolargo:
Top !

Thanks @RazCrimson !

nicolargo:
For a minor change, it is better to accept the changes and make the fix cleaner later when possible...

For a big PR, you can ask to the contributor to rework.

rhoriguchi:
thx @RazCrimson

RazCrimson:
> thx @RazCrimson 

@rhoriguchi 
Sorry about the comments on the issue. 

I'm new to reviewing PRs here. 😅

If you are interested in contributing more, feel free to take a look at other open issues.

Issue #2337, opened 2023-04-08
Tabs in process command line args corrupt the display


amurzeau:
**Describe the bug**

When a process contains a new line within its command line, glances display is corrupted.
This happens with vscode remote processes, vscode use new lines in processes command lines.

For example, see here in white rectangles:

![image](https://user-images.githubusercontent.com/5435069/230746005-900b4b31-2485-46ed-bef5-846c2170eafc.png)

Topmost rectangle: `trap "exit 0" 15` and `exec $0` are from the docker container command line (which starts with `/bin/sh -c echo Container started`)

Bottommost rectangle: `const fs = require('fs` is from `node -e` process on the previous line.

**To Reproduce**
1. Start Glances without any options
3. Start this command from a console (keep tabs and new lines, use copy paste to ensure they are kept):
```
sh -c '
while true; do true; done
						# const net = require('net');
'
```
5. In Glances, order processes by CPU consumption
6. See the sh process with `# const net = require('net');` corrupting the Glances display

Note: when the command line doesn't have tabs, no display corruption occurs (new lines are replaced by spaces).

**Expected behavior**
The new line characters in the process command line should be replaced with a space to ensure there is no corruption of the display.

**Desktop (please complete the following information):**
 - OS: Debian Unstable (Debian GNU/Linux 12 (bookworm))
 - Glances Version: Glances v3.3.1.1 with PsUtil v5.9.4
 - Glances logs file
```
# tail -20 /root/.local/share/glances/glances.log
2023-04-09 01:24:52,029 -- WARNING -- Missing Python Lib (No module named 'wifi'), Wifi plugin is disabled
2023-04-09 01:24:52,029 -- WARNING -- Wifi lib is not compliant with Python 3, Wifi plugin is disabled
2023-04-09 01:24:52,152 -- WARNING -- Sparklines module not found (No module named 'sparklines')
2023-04-09 01:24:59,864 -- INFO -- Stop Glances (keypressed: 113)
2023-04-09 01:26:18,813 -- INFO -- Start Glances 3.3.1.1
2023-04-09 01:26:18,813 -- INFO -- CPython 3.11.2 (/usr/bin/python3) and psutil 5.9.4 detected
2023-04-09 01:26:18,815 -- INFO -- Read configuration file '/etc/glances/glances.conf'
2023-04-09 01:26:18,820 -- INFO -- Start GlancesStandalone mode
[...]

PR #2349, opened 2023-04-19
Handle tabs in process arguments

RazCrimson:
#### Description

Handling line endings and tab characters to not break UI

#### Resume

* Bug fix: yes
Fixes: #2337 

@nicolargo 
How about escaping `\n` to `\\n` (similarly for `\r\n` and `\t`) instead of replacing them with a space so that they will get displayed on the UI? 

One issue with this is that we could get processes with just line endings like:
`executable\narg1\narg2\narg3`

But we could maybe add a space along with the escapes to make it look like:
`executable \narg1 \narg2 \narg3`

What are your thoughts on this?


nicolargo:
Thanks for the patch @RazCrimson !

PR #2376, opened 2023-05-08
Run CI tests on Windows with Github Actions

RazCrimson:
#### Description

Make the windows CI live again with Github Actions. 

Also removes the files for appveyor as we are practically just ignoring it now.

@nicolargo 
I think we should make use of isort, ruff, precommit and other tools and add some stringent checks for them into the CI along with glances v4. What are your thoughts on this?

#### Resume

* Bug fix: no
* New feature: yes


RazCrimson:
@nicolargo 

pre-commit is a nice tool to manage git commit hooks to automatically run tools when making commits, pushing changes.
Ref: https://pre-commit.com/

I am thinking of a pipeline that works as follows:
```
ruff(linting) -> isort(sort imports) -> black(code format) -> testing on python versions (windows & linux)
```

In the Github Actions, the tools will be run in `check` mode, where they will fail (non zero exit-code) if running them in `fix` mode would causes any changes to the source code.

In our local host, pre-commit would setup githooks to automatically run the tools in `fix` mode to rectify any possible corrections at commit-time.

precommit takes care of running the linters locally to reduce manually burden involved.
CI pipeline checks if the linters were run on the commits made.

Note: Planning this for py3.8+ as this would be released with glancesv4

Issue #1375, opened 2018-12-19
How to use GPU stats under a docker container?


oleksandrlazariev:
Hi! I could not find how to enable GPU monitoring when Glances is running under a docker container using the official docker image and the `docker run -d --name glances --restart="always" -p 61208-61209:61208-61209 -e GLANCES_OPT="-w" -v /var/run/docker.sock:/var/run/docker.sock:ro --pid host docker.io/nicolargo/glances`
I've also entered into the running container to check if `nvidia-ml-py3` installed, and it is. 
Is it possible to monitor GPU stats in this way at all? What should I do if yes?

nicolargo:
Never tested it inside a container... Perhaps you should have a look on the official **nvidia-ml-py**  lib and check how the stats are grabbed from the system.

bradcornford:
I've just tackled this myself, using docker compose and nvidia-container-runtime (https://github.com/NVIDIA/nvidia-container-runtime), i used the following: 

`
version: "2.3"

services:
  glances:
    image: nicolargo/glances
    container_name: glances
    pid: host
    network_mode: host
    privileged: true
    runtime: nvidia
    environment:
      - GLANCES_OPT=-w -6
      - NVIDIA_VISIBLE_DEVICES=all
      - NVIDIA_DRIVER_CAPABILITIES=compute,video,utility
    volumes:
      - /var/run/docker.sock:/var/run/docker.sock:ro
    restart: unless-stopped
`

rubens-curious:
Didnt work for me, here is, is think, a more complete answer : 
There is two missing dependencies for it to work (atleast for nvidia gpu).

Here is the Dockerfile : 
`
FROM nicolargo/glances
RUN pip install py3nvml
RUN pip install nvidia-ml-py3
CMD python3 -m glances -C /etc/glances.conf $GLANCES_OPT
`

Which I put in a folder named "glances" right next to the docker-compose.yml file.

Here is the docker-compose service : 
`
version: '3.8'

services:
 glances:
    build: "./glances"
    container_name: "glances"
    pid: host
    privileged: true
    runtime: nvidia
    environment:
        - GLANCES_OPT=-w
        - NVIDIA_VISIBLE_DEVICES=all
    volumes:
        - '/var/run/docker.sock:/var/run/docker.sock:ro'
    network_mode: host
    restart: always
`

RazCrimson:
@landcraft 
Intel GPUs are not supported by Glances 
[...]

PR #2386, opened 2023-05-11
chg: Dockerfile - structured & cleaner build process

RazCrimson:
@nicolargo 

I've tried to make the docker builds more cleaner. 
PTAL. Changes still need to be reviewed.

Couple of Questions
1. Why is are the non-dev tags being built off the glances being pulled from pypi?
2. Why is the nvidia/cuda image used for the ubuntu images?
3. Why was a custom pip mirror used for the ubuntu builds? (mirrors.aliyun.com)

Notable difference from the original process:
- Build dependencies are no longer present in final images
- Glances is installed directly from the source code rather than from pypi
- glances built and run on a venv installed on the container

I've pushed the built images for dev on dockerhub for ref: https://hub.docker.com/r/razcrimson/glances/tags

The images seem to be running fine for now. Not all plugins were tested, but there shouldn't be any big issues.
Results seems very good. Here is a small comparison:

| Tag        | Old Size | New Size |
|------------|----------|----------|
| dev        | 1.3 GB   | 163 MB   |
| ubuntu-dev | 760 MB   | 336 MB   |
| latest     | 74 MB    | 69 MB    | 

Sizes were measured using dive: https://github.com/wagoodman/dive

### DIFF
--- a/.dockerignore
+++ b/.dockerignore
@@ -1,2 +1,16 @@
-.dockerignore
-.git
+# Ignore everything
+*
+
+# Include only code files
+!/glances/**/*.py
+
+# Include WebUI files (remove when webui moved to seperate package)
+!/glances/outputs/static
+
+# Include Requirements files
+!/requirements.txt
+!/webui-requirements.txt
+!/optional-requirements.txt
+
+# Include Config file
+!/docker-compose/glances.conf
--- a/docker-files/alpine.Dockerfile
+++ b/docker-files/alpine.Dockerfile
@@ -50,7 +50,7 @@ RUN apk add --no-cache \
   libffi-dev \
   openssl-dev
 
-RUN python${PYTHON_VERSION} -m venv venv
+RUN python${PYTHON_VERSION} -m venv --system-site-packages --without-pip venv
 
 COPY requirements.txt webui-requirements.txt optional-requirements.txt ./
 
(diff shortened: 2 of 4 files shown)


nicolargo:
Very interesting !

The last majors changes regarding Docker images has been made by @Poeschl (in 2021)...

_Why is are the non-dev tags being built off the glances being pulled from pypi?_
Do not know...

_Why is the nvidia/cuda image used for the ubuntu images?_
The nvidia/cuda image includes all pre-requirements for the GPU plugin. Does it work on your side with the news images ?

_Why was a custom pip mirror used for the ubuntu builds? (mirrors.aliyun.com)_
Do not know...

RazCrimson:
@nicolargo 

> The nvidia/cuda image includes all pre-requirements for the GPU plugin. Does it work on your side with the news images ?

Yep, just had to use some docker options and install py3nvml (dep for gpu plugin) and it works.
Ref for docker options: https://github.com/nicolargo/glances/issues/1375#issuecomment-1474802818

Do you see anything in the changes that might cause issues? 

If not I'll do some basic checks on the generated images and then merge the changes into develop for glances v3.4 release.

nicolargo:
We need to update the documention and the docker-compose file with the option to enable GPU plugin. After that ok ton merge the PR in develop.

I will work on the release this week-end.

Thanks !

RazCrimson:
@nicolargo 

Please review the docs changes and merge if they good to go.

nicolargo:
Merged !

PR #2419, opened 2023-05-20
Draft: Add glances binary to '/usr/local/bin' + Update ENV PATH to include '/venv/bin' in Dockerfiles

Antiz96:
#### Description

This PR aims to add the `glances` binary back, as well as prepend `/venv/bin/` to the PATH.

#### Resume

* Bug fix: yes
* New feature: no
* Fixed tickets: #2406, #2418

PR #2447, opened 2023-06-09
Double quote array expansions to avoid re-splitting elements in docker-bin.sh

Antiz96:
#### Description

This PR aims to add double quotes around `$@` in the [docker-bin.sh](https://github.com/nicolargo/glances/blob/develop/docker-bin.sh) script (which I introduced in https://github.com/nicolargo/glances/pull/2419) to prevent globbing and word splitting.
For more information, see: https://www.shellcheck.net/wiki/SC2068

*To be fair, I can't think of a particular case where globbing and word splitting could have been an issue with `glances` but it doesn't hurt having a properly syntaxed/written script :smile:*

#### Resume

* Bug fix: yes *(kinda)*
* New feature: no

### DIFF
--- a/docker-bin.sh
+++ b/docker-bin.sh
@@ -1,3 +1,3 @@
 #!/bin/sh
 
-/venv/bin/python3 -m glances $@
+/venv/bin/python3 -m glances "$@"


nicolargo:
Thanks for the PR @Antiz96 

Merged into develop.

PR #2471, opened 2023-06-23
Support multiple docker socket

loopback-kr:
#### Description

Docker can separate docker **context** by user. That is, we can have multiple **docker.sock** files per user. The `/var/run/docker.sock` file is default socket file for default context, but `run/user/<USER-ID>/docker.sock` file also can exists for each user account, and the user can switch context and use docker in that context. eg, docker rootless mode, but i think glances binds only one `docker.sock` file through the volume option [when create a new container of glances](https://github.com/nicolargo/glances#docker-the-fun-way).

I wish glances could take multiple docker.sock files and monitor containers per context when glances.conf is set as below.
```
[containers]
# Define Podman sock
#podman_sock=unix:///run/user/1000/podman/podman.sock
# Define Docker sock
#docker_sock=unix:///var/run/docker.sock
docker_sock=unix:///run/user/1000/docker.sock,unix:///run/user/1001/docker.sock
```

I just push a draft version for CLI mode.
The web interface and Podman parts also need to be implemented.
I hope it's worth it. Thanks.

#### Resume

* Bug fix: no
* New feature: yes

### DIFF
--- a/docker-compose/glances.conf
+++ b/docker-compose/glances.conf
@@ -425,7 +425,7 @@ all=False
 # Define Podman sock
 #podman_sock=unix:///run/user/1000/podman/podman.sock
 # Define Docker sock
-#docker_sock=unix://var/run/docker.sock
+docker_sock=unix:///var/run/docker.sock
 
 [amps]
 # AMPs configuration are defined in the bottom of this file
--- a/docker-compose/glances.conf
+++ b/docker-compose/glances.conf
@@ -424,6 +424,8 @@ max_name_size=20
 all=False
 # Define Podman sock
 #podman_sock=unix:///run/user/1000/podman/podman.sock
+# Define Docker sock
+#docker_sock=unix://var/run/docker.sock
 
 [amps]
 # AMPs configuration are defined in the bottom of this file
--- a/glances/plugins/containers/engines/docker.py
+++ b/glances/plugins/containers/engines/docker.py
@@ -215,25 +215,26 @@ class DockerContainersExtension:
 
     CONTAINER_ACTIVE_STATUS = ['running', 'paused']
 
-    def __init__(self, server_url:str = 'unix://var/run/docker.sock'):
+    def __init__(self, server_urls:str|list = 'unix://var/run/docker.sock'):
         if import_docker_error_tag:
             raise Exception("Missing libs required to run Docker Extension (Containers) ")
 
-        self.client = None
+        self.clients = None
         self.ext_name = "containers (Docker)"
         self.stats_fetchers = {}
 
-        self.connect(server_url)
+        self.connect(server_urls)
 
-    def connect(self, base_url:str = 'unix://var/run/docker.sock'):
+    def connect(self, base_urls:str|list = 'unix://var/run/docker.sock'):
         """Connect to the Docker server."""
         # Init the Docker API Client
+        base_urls = [base_urls] if isinstance(base_urls, str) else base_urls
         try:
             # Do not use the timeout option (see issue #1878)
-            self.client = docker.DockerClient(base_url)
+            self.clients = [docker.DockerClient(url) for url in base_urls]
         except Exception as e:
             logger.error("{} plugin - Can't connect to Docker ({})".format(self.ext_name, e))
-            self.client = None
+            self.clients = None
 
     def update_version(self):
         # Long and not useful anymore because the information is no more displayed in UIs
@@ -248,7 +249,7 @@ class DockerContainersExtension:
     def update(self, all_tag):
         """Update Docker stats using the input method."""
 
-        if not self.client:
+        if not self.clients:
             return {}, []
 
         version_stats = self.update_version()
@@ -257,7 +258,8 @@ class DockerContainersExtension:
         try:
             # Issue #1152: Docker module doesn't export details about stopped containers
             # The Containers/all key of the configuration file should be set to True
-            containers = self.client.containers.list(all=all_tag)
+            containers = []
+            [[containers.append(container_per_client) for container_per_client in client.containers.list(all=all_tag)] for client in self.clients]
         except Exception as e:
             logger.error("{} plugin - Can't get containers list ({})".format(self.ext_name, e))
             return version_stats, []
--- a/glances/plugins/containers/engines/docker.py
+++ b/glances/plugins/containers/engines/docker.py
@@ -215,7 +215,7 @@ class DockerContainersExtension:
 
     CONTAINER_ACTIVE_STATUS = ['running', 'paused']
 
-    def __init__(self):
+    def __init__(self, server_url:str = 'unix://var/run/docker.sock'):
         if import_docker_error_tag:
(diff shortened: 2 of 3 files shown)


docker-compose/glances.conf
```diff
@@ -424,6 +424,8 @@ max_name_size=20
 all=False
 # Define Podman sock
 #podman_sock=unix:///run/user/1000/podman/podman.sock
+# Define Docker sock
+docker_sock=unix:///var/run/docker.sock
```
nicolargo, line 428: It will be better to let default value of docker_sock to empty (not defined) in order to use (by default) the standard docker.from_env() builder. 
Glances is multi-platform and i am pretty sure that the default value of docker_sock on Windows is not unix:///var/run/docker.sock.

glances/plugins/containers/engines/docker.py
```diff
@@ -215,25 +215,26 @@ class DockerContainersExtension:
 
     CONTAINER_ACTIVE_STATUS = ['running', 'paused']
 
-    def __init__(self):
+    def __init__(self, server_urls:str|list = 'unix:///var/run/docker.sock'):
```
nicolargo, line 218: Same remark, let the server_urls default value to '' (empty).

glances/plugins/containers/engines/docker.py
```diff
@@ -215,25 +215,26 @@ class DockerContainersExtension:
...
-        self.client = None
+        self.clients = None
         self.ext_name = "containers (Docker)"
         self.stats_fetchers = {}
 
-        self.connect()
+        self.connect(server_urls)
 
-    def connect(self):
+    def connect(self, base_urls:str|list = 'unix:///var/run/docker.sock'):
```
nicolargo, line 228: Idem

glances/plugins/containers/engines/docker.py
```diff
@@ -215,25 +215,26 @@ class DockerContainersExtension:
...
         """Connect to the Docker server."""
         # Init the Docker API Client
+        base_urls = [base_urls] if isinstance(base_urls, str) else base_urls
         try:
             # Do not use the timeout option (see issue #1878)
-            self.client = docker.from_env()
```
nicolargo, line 233: test if base_urls is empty then use from_env() else use the base_urls list instead

glances/plugins/containers/model.py
```diff
@@ -76,13 +76,13 @@ def __init__(self, args=None, config=None):
...
         # Init the Docker API
-        self.docker_extension = DockerContainersExtension() if not import_docker_error_tag else None
+        self.docker_extension = DockerContainersExtension(server_urls=self._docker_sock()) if not import_docker_error_tag else None
 
         # Init the Podman API
         if import_podman_error_tag:
             self.podman_client = None
         else:
-            self.podman_client = PodmanContainersExtension(podman_sock=self._podman_sock())
+            self.podman_client = PodmanContainersExtension(podman_sock=self._podman_sock()) # TODO: podman also
```
nicolargo, line 85: IS it also possible to have multiple socket for Podman ?

glances/plugins/containers/model.py
```diff
@@ -91,6 +91,16 @@ def __init__(self, args=None, config=None):
         self.update()
         self.refresh_timer.set(0)
 
+    def _docker_sock(self):
+        """Return the docker socks.
+        Default value: unix:///var/run/docker.sock
+        """
+        conf_docker_sock = self.get_conf_value('docker_sock')
+        if len(conf_docker_sock) == 0:
+            return "unix:///var/run/docker.sock"
```
nicolargo, line 100: To be refactor according to previous comments

nicolargo (changes requested):
Hi @loopback-kr  !

Thanks for the interesting PR.

For the moment, i can not merge it to the develop branch. I have added some comments inside your code.

Nicolas

docker-compose/glances.conf
RazCrimson: +1 for this

glances/plugins/containers/engines/docker.py
RazCrimson: Just something to note with Python support.

Glances v4 would be target Py3.8+

The `str |  list` syntax was only support from Py3.10 onwards, iirc.
So better use the older Union[] style to prevent older interpreters from crashing.

Also I think returning a list[str] even when there is only a single elemenet would reduce unnecessary type checks later

glances/plugins/containers/engines/docker.py
```diff
@@ -215,25 +215,26 @@ class DockerContainersExtension:
...
+            self.clients = [docker.DockerClient(url) for url in base_urls]
```
RazCrimson, line 234: This will end up failing/dropping all socket connections if any one of them fail.

I think a more tolerant approach where we just ignore the connections that fail would be better as they user will still get to see some of the containers if they are active.

A log message indicating the failed socket would be helpful for debugging

glances/plugins/containers/engines/docker.py
```diff
@@ -257,7 +258,8 @@ def update(self, all_tag):
         try:
             # Issue #1152: Docker module doesn't export details about stopped containers
             # The Containers/all key of the configuration file should be set to True
-            containers = self.client.containers.list(all=all_tag)
+            containers = []
+            [[containers.append(container_per_client) for container_per_client in client.containers.list(all=all_tag)] for client in self.clients]
```
RazCrimson, line 262: Same problem as mentioned earlier, if one of the socket connections break or the server ends up serving a malformed response, then we will end up with no results.

glances/plugins/containers/model.py
RazCrimson: @nicolargo 
Yep.
In fact its more common to have custom socket's in the podman ecosystem.

glances/plugins/containers/model.py
```diff
@@ -297,6 +307,8 @@ def msg_curse(self, args=None, max_width=None):
         ret.append(self.curse_add_line(msg))
         msg = ' {:<7}'.format('Tx/s')
         ret.append(self.curse_add_line(msg))
+        msg = '{:<30}'.format('Socket URL')
```
RazCrimson, line 310: We also should display this only optionally. As in the user can opt in if they want this to be displayed.

The containers plugin has very little screen room, so this will reduce it even further

RazCrimson:
@loopback-kr 
I'll take care of the clean up and merge, if you don't mind.

Thanks for the contribution!

loopback-kr:
There is not enough time for this work. I would be happy if you could complete the code for me.

nicolargo:
@RazCrimson Do you think that this PR can be rejected ?

RazCrimson:
> @RazCrimson Do you think that this PR can be rejected ?

I'll go ahead fix up the changes and update the PR by 4.1.0

thorian93:
Arriving almost two years late to the party, but I would really love to see this feature happen.
But I do not know the first thing about coding. Can anyone tell, if this might make it to an official release soon?

Issue #2439, opened 2023-06-01
incorrect linux_distro in docker version glances


s0urcelab:
**Describe the bug**
i use docker version glances on my Debian PC, ```system.linux_distro``` shows 'Alpine Linux 3.17.1'

**To Reproduce**
web api ```/api/3/all``` field ```system.linux_distro```

**Expected behavior**
should give 'Debian Linux xxxx'

**Screenshots**

**Environement (please complete the following information)**
 - Operating System (lsb_release -a or OS name/version):  Debian GNU/Linux 11 (bullseye)
 - Glances & psutil versions: `To be completed with result of: nicolargo/glances:latest-full
 - How do you install Glances (Pypi package, script, package manager, source): docker
 - Glances test (only available with Glances 3.1.7 or higher):

nicolargo:
When using the Glances Docker, it runs on "Alpine Linux" OS. If you want to fake the information and show the host operating system you can add the following option to the docker exec command line: **-v /etc/os-release:/etc/os-release:ro**

s0urcelab:
> When using the Glances Docker, it runs on "Alpine Linux" OS. If you want to fake the information and show the host operating system you can add the following option to the docker exec command line: **-v /etc/os-release:/etc/os-release:ro**

mount volume works, but i think this behavior is a little weird.

RazCrimson:
@s0urcelab 
This would be the expected behavior if you run official glances images through docker. 

Maybe we could add this volume mount detail in the docker readme file?

s0urcelab:
@RazCrimson Good advice! It would be helpful if we can add this description in document.

PR #2473, opened 2023-06-24
Docs: docker - include OS Release Volume mount info

RazCrimson:
Closes: #2439

Include the volume mounting info for OS Release to get glances on docker displaying host info.
 
Presenting the image variants in a table as I think it would make it easier to compare them.


nicolargo:
Thanks @RazCrimson !

PR #2505, opened 2023-07-20
Readd support for arm/v6 to docker image

claha:
The support for arm/v6 was dropped since the build was based on python, which did not have support for arm/v6, but now they are based on alpine.

#### Description

I tried running glances using docker on my raaspberry pi zero but it wouldn't start. Digged through some old issues and found that support for arm/v6 was dropped some time ago. But I think it should be fine to add it back now.O

#### Resume

* Bug fix: no
* New feature: yes


claha:
Seems the api key stored in the repo has changed during that last day? The logs say:

Warning:  It looks like you are trying to use an API token to authenticate in the package index and your token value does not start with "pypi-" as it typically should. This may cause an authentication error. Please verify that you have copied your token properly if such an error occurs.

This warning didn't come up for the other PRs which were created a few days before mine.

nicolargo:
Thanks for the PR @claha  !

nicolargo:
image

Image available in Docker hub.

PR #2559, opened 2023-09-26
Update README.rst

evrial:
Simpler method to install psutil and get glances running on Debian/Ubuntu.


nicolargo:
Thanks for the PR @evrial !

PR #2563, opened 2023-09-30
Update docker.rst to remove latest-alpine tag

NLZ:
#### Description

`latest-alpine` tag is decrepated, so updating the compose example to use `latest` instead

#### Resume

* Bug fix: no
* New feature: no


nicolargo:
Thanks for the PR @NLZ !

Issue #1738, opened 2020-09-26
No module named 'influxdb' on the snap version of glances


gbonazzoli:
#### Description

I'm using the snap version of glances on Oracle Linux 8.2, in order to keep the operating system as clean as possible from package installation.

glances goes like a charm but I can not load an influxdb database withe the metrics collected.

```
[root@dblinux 902]# glances -C/root/glances.conf --export influxdb
Traceback (most recent call last):
  File "/snap/glances/902/bin/glances", line 33, in <module>
    sys.exit(load_entry_point('Glances==3.1.5', 'console_scripts', 'glances')())
  File "/snap/glances/902/lib/python3.5/site-packages/glances/__init__.py", line 143, in main
    start(config=config, args=args)
  File "/snap/glances/902/lib/python3.5/site-packages/glances/__init__.py", line 108, in start
    mode = GlancesMode(config=config, args=args)
  File "/snap/glances/902/lib/python3.5/site-packages/glances/standalone.py", line 51, in __init__
    self.stats = GlancesStats(config=config, args=args)
  File "/snap/glances/902/lib/python3.5/site-packages/glances/stats.py", line 48, in __init__
    self.load_modules(self.args)
  File "/snap/glances/902/lib/python3.5/site-packages/glances/stats.py", line 103, in load_modules
    self.load_exports(args=args)
  File "/snap/glances/902/lib/python3.5/site-packages/glances/stats.py", line 178, in load_exports
    export_module = __import__(self._exports_all[export_name])
  File "/snap/glances/902/lib/python3.5/site-packages/glances/exports/glances_influxdb.py", line 27, in <module>
    from influxdb import InfluxDBClient
ImportError: No module named 'influxdb'
```

I think it is only a problem on how the snap package is fullfilled.

Any chance to have an official snap with the functionality on board ?

#### Versions
```
[root@dblinux 902]# snap list
Name      Version        Rev    Tracking       Publisher    Notes
glances   3.1.5          902    latest/stable  nicolargo    -

[root@dblinux 902]# glances -V
Glances v3.1.5 with PsUtil v5.7.2
Log file: /root/snap/glances/902/.cache/glances/glances.log
```

nicolargo:
@Lin-Buo-Ren What do you think of this request ? Is there a simple and good way to include Glances extra module in the default Snap ? Or is it better to create a new "Glances-Full" Snap ?

Lin-Buo-Ren:
I'd say we bundle all deps in one package.

[...]

PR #2669, opened 2024-01-29
Add missing dependencies for influxdb2 and web

cmatsuoka:
#### Description

Running as a web server or exporting to influxdb 2.x depends on the bottles and influxdb-client modules to work correctly. This fixes the missing package errors in the snap package.

A similar approach can probably be used to address #1738.

**Note:** this targets `support/glancesv3` as the current develop head fails to initialize the terminal on my system and could not be tested. However, the changes are trivial and can be easily be used as a reference to redo it against the `develop` branch.

#### Resume

* Bug fix: yes
* New feature: no
* Fixed tickets:

### DIFF
--- a/requirements.txt
+++ b/requirements.txt
@@ -6,3 +6,4 @@ ujson<3; python_version < "3.0"
 ujson<4; python_version >= "3.5" and python_version < "3.6"
 ujson<5; python_version >= "3.6" and python_version < "3.7"
 ujson>=5.4.0; python_version >= "3.7"
+influxdb==1.39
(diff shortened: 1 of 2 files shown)


nicolargo:
Hi @cmatsuoka 

The requirements.txt and setup.py file are not only used to generate the Snap package but also for all the other packaging (Pypi). We want to avoid non mandatory dependencies in those packages.

So we had to find a way to only include additional requirements in the Snap package.

Any any inpout are welcome.

Nicolas

Issue #2181, opened 2022-11-05
Replace Bottle by FastAPI


nicolargo:
Trade-off should be made between the current Bottle Framework and a new one like FastAPI (see https://fastapi.tiangolo.com/#requirements).

RazCrimson:
Another possible option to consider: 
https://github.com/starlite-api/starlite

nicolargo:
Should be linked to https://github.com/nicolargo/glances/issues/2114

Tronic:
Consider https://sanic.dev/en/ - one of the fastest Python frameworks (talking about 100 kreq/s) and has API very similar to FastAPI or Flask+async. Already proven, used in large projects (e.g. Italian government Covid tracing app). Can directly run a public web server incl. HTTP/3, and comes with other advanced features while staying low level and unopionated. Just almost nobody seems to know it exists. Disclaimer: I am a developer of it.

RazCrimson:
@Tronic 

Assuming that the FastAPI are not biased, it might be better to just stick with FastAPI
Ref: https://fastapi.tiangolo.com/alternatives/#sanic


Litestar/Starlite seemed interesting but seems to be quite less used than FastAPI. FastAPI also has a huge userbase with lot of community support, so I think sticking with it would be the best bet in a long term standpoint. 

@nicolargo What do you think?

Tronic:
@RazCrimson Starlette is an ASGI server and can be used with Sanic as well. However, the preferred way to run Sanic is using its integrated HTTP server which runs much faster than what is possible via ASGI. That being said, FastAPI can run fairly quick too. I've yet to see useful conclusive benchmarking between the two, and we do not intend to compete with it for users (rather, we are interested on how to awake Flask and Django users, as well as NodeJS Express users, to that there are modern alternatives).

RazCrimson:
@Tronic 

My bad, I should have put that a bit better.

The reason I preferred FastAPI to Sanic was not really the performance but the huge community behind it. So in the end FastAPI would end up being more tested for bugs and vulnerabilities than other frameworks as the larger user base would be having more diverse use cases.

Tronic:
@RazCrimson Not sure whether you'll get that, as it is a very young framework and only built for JSON APIs (rather than diverse use cases). But yes, it certainly seems more popular at this time and has more active developers, so I understand why you'd go with that.

nicolargo:
Work started on branch issue2181

nicolargo:
Merged into develop.

[...]

PR #2672, opened 2024-02-03
Add bottle and influxdb python packages to snap

cmatsuoka:
#### Description

Add optional packages to allow glances run as a server or export to
influxdb2 when executed from the snap package.

This affects only snap packaging and won't add optional packages to
the default installation.

This patch can be directly applied to `support/glancesv3` to fix the
current published snap. It was not tested with the `develop` branch
due to failure to initialize the terminal.

#### Resume

* Bug fix: yes
* New feature: no
* Fixed tickets:

### DIFF
--- a/snap/snapcraft.yaml
+++ b/snap/snapcraft.yaml
@@ -45,11 +45,16 @@ parts:
     - selective-checkout
     plugin: python
     source: .
+    python-packages:
+      - bottle
+      - influxdb-client==1.39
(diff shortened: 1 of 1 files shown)


cmatsuoka:
This should be reviewed by @brlin-tw.

nicolargo:
Hi @cmatsuoka 

Thanks for the PR.

Bottle is not used anymore in the develop branch. It has been replaced by FastAPI (see https://github.com/nicolargo/glances/issues/2181).

Issue #2387, opened 2023-05-11
better support for white / light terminal background


j-lakeman:
**Is your feature request related to a problem? Please describe.**
glances should be executed with the `--theme-white` flag on white / light terminal backgrounds automatically.

**Describe the solution you'd like**
glances needs to autodetect the current terminal theme

**Describe alternatives you've considered**
tinkering with aliases, doesn't work if terminal theme changes while glances is running though

nicolargo:
There is no standard way to detect the terminal background. There are a few ways that work on some terminals, but on most terminals it is impossible AFAIK.

nicolargo:
Some implementation exist, but not work on standard Gnone terminal, terminator... 

https://github.com/rocky/shell-term-background

j-lakeman:
What about adopting to the current OS theme? Would that be easier? That would solve my issue as well.

nicolargo:
@j-lakeman do not find any Pythonific way to get the OS theme... Any idea and tips are welcome.

j-lakeman:
https://github.com/albertosottile/darkdetect
Does that help?

nicolargo:
Darkdetect only work for Gnome theme and not for the terminal background color :(

No simple and cross platform way to retrieve terminal background: https://stackoverflow.com/a/78629028/1919431

PR #2687, opened 2024-03-01
Use normal color for normal text instead of an arbitrary color

turbocrime:
#### Description

This makes glances use the normal terminal color for most text, instead of an explicit 'black' or 'white' color that may not be visible.

* Bug fix: yes
* New feature: yes
* Fixed tickets: #2387 

This even supports dynamic changing terminal colorschemes without any restart, like a terminal switching from 'light' to 'dark' mode at sunset.

### DIFF
--- a/conf/glances.conf
+++ b/conf/glances.conf
@@ -23,8 +23,6 @@ history_size=1200
 ##############################################################################
 
 [outputs]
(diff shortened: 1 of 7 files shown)


nicolargo:
Thanks for the PR @turbocrime !

Merged into develop.

Will be release in Glances 4.0.

PR #2699, opened 2024-03-13
Add CallbackAPIversion for MQTT export

erdnaxela02:
#### Description

Due to the recent update of paho-mqtt to version V2.0.0 which requires API versioning, I added callback_api_version VERSION2 in paho.Client to warn of future removal of VERSION1

When installing paho-mqtt with the command 'pip install paho-mqtt' this automatically installs the latest version (currently 2.0.0)
This caused an error when launching Glances with the `--export mqtt` option.
`Connection to MQTT server xxx.xxx.xxx.xxx:XXXX failed with error: Client.__init__() missing 1 required positional argument: 'callback_api_version'`
The workaround was to install paho-mqtt V1.X with this command 'pip install "paho-mqtt<2"

With this fix, Glances works with the new version of paho-mqtt

#### Resume

* Bug fix: yes
* New feature: no
* Fixed tickets: not for the moment

### DIFF
--- a/docker-compose/glances.conf
+++ b/docker-compose/glances.conf
@@ -595,6 +595,7 @@ queue=glances_queue
 # Configuration for the --export mqtt option
 host=localhost
 port=8883
+hostname=NONE
 tls=false
 user=guest
 password=guest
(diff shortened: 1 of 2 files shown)


erdnaxela02:
I'm closing this PR because I think I made a mistake. I'm not sure how to use github between fork and branches

PR #2701, opened 2024-03-15
Add MQTT hostname configuration

erdnaxela02:
#### Description

I added the possibility of configuring the hostname of the machine in the MQTT topic.
Are you going to ask me why?
Well I have a dualboot machine. Two different operating systems but the same hardware.
But the 'socket.gethostname()' command returns a different value depending on the operating system ("DESKTOP-B1JV0TR" on Windows and "My-ubuntu" on Linux)
This has the consequence that the access path to the topic is different in my broker even though it is the same machine.
However, I need to read the information on a topic without having to modify it depending on the OS on which I started my machine.
I thought this might be useful to other people.

#### Resume

* Bug fix: no
* New feature: yes
* Fixed tickets: not for the moment

### DIFF
--- a/docker-compose/glances.conf
+++ b/docker-compose/glances.conf
@@ -595,7 +595,6 @@ queue=glances_queue
 # Configuration for the --export mqtt option
 host=localhost
 port=8883
-hostname=NONE
 tls=false
 user=guest
 password=guest
--- a/docker-compose/glances.conf
+++ b/docker-compose/glances.conf
@@ -595,6 +595,7 @@ queue=glances_queue
 # Configuration for the --export mqtt option
 host=localhost
 port=8883
+hostname=NONE
 tls=false
 user=guest
 password=guest
--- a/glances/exports/mqtt/__init__.py
+++ b/glances/exports/mqtt/__init__.py
@@ -38,13 +38,17 @@ class Export(GlancesExport):
 
         # Load the MQTT configuration file
         self.export_enable = self.load_conf(
-            'mqtt', mandatories=['host', 'password'], options=['port', 'user', 'topic', 'tls', 'topic_structure']
+            'mqtt', mandatories=['host', 'password'], options=['port', 'user', 'hostname', 'topic', 'tls', 'topic_structure']
         )
         if not self.export_enable:
             exit('Missing MQTT config')
 
         # Get the current hostname
-        self.hostname = socket.gethostname()
+        self.hostname = (self.hostname or socket.gethostname())
+        if self.hostname in ['NONE']:
+            self.hostname = socket.gethostname()
+        else:
+            self.hostname = self.hostname
         self.port = int(self.port) or 8883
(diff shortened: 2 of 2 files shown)


docker-compose/glances.conf
```diff
@@ -595,6 +595,7 @@ queue=glances_queue
 # Configuration for the --export mqtt option
 host=localhost
 port=8883
+hostname=NONE
```
nicolargo, line 598: Not necessary

glances/exports/mqtt/__init__.py
```diff
@@ -38,13 +38,17 @@ def __init__(self, config=None, args=None):
 
         # Load the MQTT configuration file
         self.export_enable = self.load_conf(
-            'mqtt', mandatories=['host', 'password'], options=['port', 'user', 'topic', 'tls', 'topic_structure']
+            'mqtt', mandatories=['host', 'password'], options=['port', 'user', 'hostname', 'topic', 'tls', 'topic_structure']
```
nicolargo, line 41: Variable name is confusing with host

glances/exports/mqtt/__init__.py
```diff
@@ -38,13 +38,17 @@ def __init__(self, config=None, args=None):
...
         if not self.export_enable:
             exit('Missing MQTT config')
 
         # Get the current hostname
-        self.hostname = socket.gethostname()
+        self.hostname = (self.hostname or socket.gethostname())
+        if self.hostname in ['NONE']:
+            self.hostname = socket.gethostname()
+        else:
+            self.hostname = self.hostname
```
nicolargo, line 51: All the line added should be replaced by:

self.hostname = self.hostname or socket.gethostname()

nicolargo (changes requested):
Hi @erdnaxela02 thanks for the PR, some comments to take into account before merging into develop

docker-compose/glances.conf
erdnaxela02: I Yes I know, but I told myself that it was necessary to put it to show as an example to other users that it was available

glances/exports/mqtt/__init__.py
erdnaxela02: thank you for this advice

glances/exports/mqtt/__init__.py
erdnaxela02: okay, what do you think of 'devicename'?
I chose hostname because this variable is already used in __init__.py

erdnaxela02:
Thank you for your advice. Excuse me if I'm not doing things right, I'm not a developer. I don't know the language. I only taught myself how to make these few modifications.

nicolargo (approved):
Also perfect !

Do not apology, any contribution are welcome in open-source !

nicolargo:
Merged into develop. It will be release in Glances 4.0.

Thanks !

PR #2712, opened 2024-03-29
Use enum instead of int for callback API version.

mpkossen:
#### Description

This fixes an issue with #2700 which results in an unsupported operand type error because an int is passed whereas paho expects an enum.

#### Resume

* Bug fix: yes
* New feature: no
* Fixed tickets: 2700


nicolargo:
Thanks @mpkossen  !

PR #2745, opened 2024-04-29
Make docker memory usage match `docker stats`

RazCrimson:
Fixes 2637

Web UI has issues


RazCrimson:
> Tested, also with the WebUI, it looks ok for me.

I just noticed that the Terminal UI displays values at Kibi/Mebi/Gibi bytes but the Web UI seems to be using the Kilo/Mega/Giga standard.

Ref: https://danielmiessler.com/p/the-difference-between-kilobytes-and-kibibytes

RazCrimson:
> Only one question: do you think that the 'old' memory_usage information can be useful ? If the reponse is yes, perhaps it will be better to have mamory_usage and memory_usage_no_cache in the data model and display by default the memory_usage_no_cache in Glances UI.
>
> What do you think ?

@nicolargo 
Now, this makes me ponder a bit about what glances is or should be.

If the vision of glances being the eye on a system includes exporting detailed stats like [prometheus](https://prometheus.io/) then we should probably export even the additional fields from docker/podman API through the export and API but just display plain usage_with_no_cache in the UI.

If we just want glances to be a quick tool to check the current stats then, we could just remove the old field and just have the fields required for the UI in the data model.

What are your thoughts on it?

nicolargo:
> > Only one question: do you think that the 'old' memory_usage information can be useful ? If the reponse is yes, perhaps it will be better to have mamory_usage and memory_usage_no_cache in the data model and display by default the memory_usage_no_cache in Glances UI.
> > What do you think ?
> 
> @nicolargo Now, this makes me ponder a bit about what glances is or should be.
> 
> If the vision of glances being the eye on a system includes exporting detailed stats like [prometheus](https://prometheus.io/) then we should probably export even the additional fields from docker/podman API through the export and API but just display plain usage_with_no_cache in the UI.
> 
> If we just want glances to be just a quick tool to check the current stats then, we could just remove the old field and just have the fields required for the UI in the data model.
> 
> What are your thoughts on it?

My vision of Glances  states closer to the your first vision: have detailed information available through the API and have a simple as possible "eye on a system" through the Curses and WebUI.

RazCrimson:
@nicolargo 
Please go ahead and merge these fixes for the Glances v4 release.

I'll tackle exposing the all the stats in #2746

nicolargo:
Done !

PR #2778, opened 2024-05-16
chg: event_list - enable data validation only if pydantic available

### LINKED ISSUE #2777: Please make `pydantic` optional dependency, not required one
@nicolargo Please consider making `pydantic` an optional dependency: forcing a dependency on Rust is undesirable for several reasons (it is a very heavy dependency to build, it is broken on some platforms and cannot be built there at all, it is questionable from security point of view, since it requires itself to build).

RazCrimson:
Closes #2777

### DIFF
--- a/glances/event.py
+++ b/glances/event.py
@@ -32,7 +32,13 @@ Item (or event) is defined by:
     }
 """
 
-from pydantic.dataclasses import dataclass
+from glances.logger import logger
+
+try:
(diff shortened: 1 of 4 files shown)


RazCrimson:
@barracuda156
Seems to work fine for me.

Could you give this a spin and let us know if everything works fine on your end?

barracuda156:
@RazCrimson Thank you! It works now.
glances

Issue #2801, opened 2024-05-26
Reduce code complexity


nicolargo:
Reduce complexity of function higher than 20:

https://scrutinizer-ci.com/g/nicolargo/glances/code-structure/develop?elementType=py-function&orderField=conditional_complexity&order=desc&changesExpanded=0

ariel-anieli:
Hello @nicolargo,
I will take on [glances.plugins.system.PluginModel.update()](https://scrutinizer-ci.com/g/nicolargo/glances/code-structure/develop/py-function/glances.plugins.system.PluginModel.update). For us to agree on the expectation; I should follow _How to fix,_ and that's it. Is it right?

nicolargo:
@ariel-anieli yes and all code optimization that loo, interesting to you .

Thanks !

ariel-anieli:
> @ariel-anieli yes and all code optimization that loo, interesting to you .
> 
> Thanks !

:slightly_smiling_face: _Je t'en prie!_ I will keep you posted.

ariel-anieli:
@nicolargo, I will take on [glances.plugins.percpu.PluginModel.msg_curse()](https://scrutinizer-ci.com/g/nicolargo/glances/code-structure/develop/py-function/glances.plugins.percpu.PluginModel.msg_curse).

nicolargo:
Huge task  :) This method is historical...

ariel-anieli:
> Huge task :) This method is historical...

:slightly_smiling_face: I'll have a try. Are there caveats worth noticing?

ariel-anieli:
@nicolargo, I am taking on [unittest-core.TestGlances._common_plugin_tests()](https://scrutinizer-ci.com/g/nicolargo/glances/code-structure/develop/py-function/unittest-core.TestGlances._common_plugin_tests).

ariel-anieli:
I will tackle [`_GlancesCurses.__catch_key()`](https://scrutinizer-ci.com/g/nicolargo/glances/code-structure/develop/py-function/glances.outputs.glances_curses._GlancesCurses.__catch_key)

ariel-anieli:
I'll take on [`PluginModel.__msg_curse_extended_process_thread()`](https://scrutinizer-ci.com/g/nicolargo/glances/code-structure/develop/py-function/glances.plugins.processlist.PluginModel.__msg_curse_extended_process_thread)

ariel-anieli:
[...]

PR #2884, opened 2024-07-22
Reduced complexity of `glances.plugins.system.PluginModel.update()`

ariel-anieli:
#### Description

Part of #2801. Local tests are all passing:

```shell
# bin/python unittest-core.py | grep -iP 'OK:fail'
.....................................
----------------------------------------------------------------------
Ran 37 tests in 13.245s

OK
```

#### Resume

Broke `update` in two parts:

* if `local`, stats taken from `get_stats_from_std_sys_lib`
* and readability achieved by `add_human_readable_name`
* else if `snmp`, taken from `update_stats_with_snmp`.

`get_stats_from_std_sys_lib` has two helpers:

1. `get_win_version_and_platform`.
2. `get_linux_version_and_distro`.

### DIFF
--- a/glances/plugins/system/__init__.py
+++ b/glances/plugins/system/__init__.py
@@ -128,6 +128,84 @@ class PluginModel(GlancesPluginModel):
         # Get the default message (if defined)
         self.system_info_msg = config.get_value('system', 'system_info_msg') if config else None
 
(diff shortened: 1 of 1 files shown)


nicolargo:
Thanks for the PR @ariel-anieli 

Merged into develop.

PR #2903, opened 2024-07-30
chg: curses - change colors for some LOG styles

### LINKED ISSUE #2882: Make fields labelled in Green easier to see
image
Currently, glances will highlighted some fields in green, but use dark grey text. This makes the text in the box rather difficult to see at a quick glance, compared to the either the boxes highlighted in red, or ones that have no highlighting at all.

Ideally, the text colour would be changed from grey to something that has better contrast when highlighted in green, such as black, which would make it stand out more compared to the white text on red.

**Describe alternatives you've considered**
One alternative I considered was to remove the green box entirely, and just label the text green, but that could be more confusing, since it would be mixed with the other text that is coloured green.

**Additional context**
image

I've included an example testing other colours (in MS Paint). Either black or white would be more legible than the original grey.

RazCrimson:
fixes #2882

### DIFF
--- a/glances/outputs/glances_colors.py
+++ b/glances/outputs/glances_colors.py
@@ -66,9 +66,9 @@ class GlancesColors:
             curses.init_pair(3, curses.COLOR_GREEN, -1)
             curses.init_pair(5, curses.COLOR_MAGENTA, -1)
         else:
-            curses.init_pair(2, -1, curses.COLOR_RED)
-            curses.init_pair(3, curses.COLOR_BLACK, curses.COLOR_GREEN)
-            curses.init_pair(5, -1, curses.COLOR_MAGENTA)
+            curses.init_pair(2, curses.COLOR_WHITE, curses.COLOR_RED)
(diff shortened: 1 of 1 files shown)


nicolargo (changes requested):
Glances should be compliant with both white and  black background.
-1 index allows to dynamicaly adapt the color to the terminal theme.

RazCrimson:
@nicolargo 

Yea, you are right. But doesn't that apply only when we are changing the background color?

Here is a sample in a white themed background:
image

Versions (Black) - Old vs New:
image
image

Versions (White) - Old vs New:
image
image

I feel using white foreground text on a colored background looks more prominent on both white and black backgrounds.

Should we  stick with the opposite (-1) of terminal background for now or move ahead with WHITE fg text?
What do you think?

nicolargo:
On my side with a black background the OK theme is more readable with a dark font than a white one.

RazCrimson:
Ok I'll go ahead and revert this for now then

Issue #2906, opened 2024-08-05
Reduce complexity of `Makefile`


ariel-anieli:
Hello,

This issue is for tracking the ongoing effort on `Makefile`. Might help for #2583:

- [x] Factorize with variables (#2907)
- [x] When targets have same templates, for instance `test-*`, use pattern rules:

1. #2910 
2. #2913
3. #2917

- [x] ~~put `tests-*` in a test folder~~ (TODO: this breaks `glances` imports in `unitttest-*.py`)
- [x] ~~evaluate if virtual paths would be helpful in the rules using docker files and test scripts~~ (TODO: first of all, `unitttest-*.py` import breaks need to be solved)
- [x] ~~Remove dependencies to external files; i.e., `*-requirements.txt`~~ (TODO: should be part a task in itself, could be extended to the Docker requirements)
- [x] ~~use patterns as in #2917. Here is the pattern: `VIRTUAL_ENV/python FLAGS TARGET`~~ (decrease legibility)

- [x] Fix bug on `make test-restful` (unrelated to `Makefile`, opened #2918)
```shell
# make test-restful 1>/dev/null
INFO:     Started server process [812050]
INFO:     Waiting for application startup.
INFO:     Application startup complete.
INFO:     Uvicorn running on http://0.0.0.0:61234 (Press CTRL+C to quit)
...F.............INFO:     Shutting down
INFO:     Waiting for application shutdown.
INFO:     Application shutdown complete.
INFO:     Finished server process [812050]
.
======================================================================
FAIL: test_003_plugins (__main__.TestGlances.test_003_plugins)
Plugins.
----------------------------------------------------------------------
Traceback (most recent call last):
  File "/home/ariel/src/python/glances/unittest-restful.py", line 120, in test_003_plugins
    self.assertIsInstance(req.json(), list)
AssertionError: {} is not an instance of <class 'list'>

----------------------------------------------------------------------
Ran 18 tests in 6.277s

FAILED (failures=1)
make: *** [Makefile:90: test-restful] Error 1
```

PR #2907, opened 2024-08-05
Refactorize `Makefile`

ariel-anieli:
#### Description

Part of #2906.

#### Resume

Factorized rules with variables.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -24,7 +24,12 @@ LASTTAG  =  $(shell git describe --tags --abbrev=0)
 .PHONY: help test docs docs-server venv venv-min venv-dev
 
 help: ## List all make commands available
-	@grep -E '^[\.a-zA-Z_%-]+:.*?## .*$$' $(MAKEFILE_LIST) | awk -F ":" '{print $1}' | grep -v % | sed 's/\\//g' | sort | awk 'BEGIN {FS = ":[^:]*?##"}; {printf "\033[1;34mmake %-50s\033[0m %s\n", $$1, $$2}'
+	@grep -E '^[\.a-zA-Z_%-]+:.*?## .*$$' $(MAKEFILE_LIST) | \
+	awk -F ":" '{print $1}' | \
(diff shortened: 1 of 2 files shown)


nicolargo:
Hi @ariel-anieli 

Thanks for the PR. Sorry for the delay, i just came back from vacation, i will have a look on it this week-end :)

ariel-anieli:
> Hi @ariel-anieli
> 
> Thanks for the PR. Sorry for the delay, i just came back from vacation, i will have a look on it this week-end :)

:slightly_smiling_face: Hi @nicolargo, _pas de soucis!_

Makefile
```diff
@@ -1,12 +1,35 @@
...
+CONF     := conf/glances.conf
+PIP      := $(VENV)/pip
+PYTHON   := $(VENV)/python
+UNITTEST := unittest
+
+DOCKER_BUILD      := docker buildx build
+DOCKER_RUN        := docker run
+DOCKERFILE_UBUNTU := docker-files/ubuntu.Dockerfile
+DOCKERFILE_ALPINE := docker-files/alpine.Dockerfile
+PODMAN_SOCK       := /run/user/1000/podman/podman.sock
```
RazCrimson, line 14: Instead of hard coding 1000, could you please use the output of `id -u` for PODMAN_SOCK.

Makefile
```diff
@@ -1,12 +1,35 @@
...
+DOCKER_SOCK       := /var/run/docker.sock
```
RazCrimson, line 15: Would recommend switching `:=` to `?=` as it would make it easy to use custom Docker sockets.

The same applies for Podman sockets too

Makefile
ariel-anieli: Thanks for the feedback; it's done.

Makefile
ariel-anieli: Done as well.

RazCrimson (approved):

PR #2924, opened 2024-08-18
Refactorized rules targetting `profiling-*` & `memory-profiling`

ariel-anieli:
#### Description

Part of #2906.

#### Resume

* `profiling-*` share the same start banner
* used `OUT_DIR` for static outputs
* `TIMES` controls the number of iterations.


nicolargo:
Merged into develop.

Thanks a lot @ariel-anieli !

PR #2997, opened 2024-11-01
Refactored `glances.plugins.ip.PluginModel.update`

ariel-anieli:
#### Description

All tests are passing:

```shell
# make test 2>&1 | grep -Pi '^ok|^fail'
OK
OK
OK
```

After:
Screenshot from 2024-11-01 15-30-40
Before:
Screenshot from 2024-11-01 15-30-54

#### Resume
* part of #2801
* formatted into a dict; if input method is SNMP, returns default stats
* if input method is local, calls `get_stats_for_local_input`
* it has three steps:  `get_private_ipv4`, `get_first_ipv4`, `get_public_ipv4`.


ariel-anieli:
Work in progress: some tests are failing.

PR #3022, opened 2024-11-17
WebUI: Fix network/diskio filter when property doesn't exist

jdrouhard:
#### Description

Most recent glances (4.2.1) was throwing a bunch of errors in the web console and not displaying the network or disk i/o sections at all. Stuff like:

```
TypeError: this.view[t.name].read_bytes_rate_per_sec is undefined
...
TypeError: this.view[t.interfaceName].bytes_recv_rate_per_sec is undefined
...
```

The error was pointing to the filter function(s) in the two plugins. If there's nothing to display, the web app doesn't populate those properties on the interface/disk objects in the view, so you can't use them in the check for whether it should be hidden or not.

#### Resume

* Bug fix: yes
* New feature: no


nicolargo:
Thanks a lot @jdrouhard !

Merged into develop.

PR #3071, opened 2024-12-31
Create a file tailer plugin

benjimons:
#### Description

This pull request is to integrate a simple file tailer into to read a file and print it to screen - useful for log files.

For any questions concerning installation or use, please open a discussion (https://github.com/nicolargo/glances/discussions), not an issue.

#### Resume

* Bug fix: no
* New feature: yes
* Fixed tickets: comma-separated list of tickets fixed by the PR, if any

PR #3078, opened 2025-01-05
Update __init__.py

benjimons:
Thanks for your feedback and kind support - here is my changes against #3071 

Disable by default maybe we can leave to the last bit.

I am not sure what to do about Windows - this is supposed to be a plain file reader so I dont know what else to do - i dont really want to get into reading event log files.

On the WebUI side of things I am a bit lost.

#### Description

Changes
1. More memory efficiency
2. Added warning for file not found
3. Condensed headers.
4. removed the 1GB colouring conditions

- [x]  The plugin code should be optimize to be usable with huge log file without any impact on the Glances performance.
- [ ]  The plugin code should be compliant with Windows OS.
- [x]  Configuration of the file size threshold should be done through the configuration file (not statically define to 1GB)
- [x]  In the TUI, space is precious... PLease optimize the first 4 lines (TITLE, File name, File size, Last line header) to two or one line
- [ ]  In the WebUI refactor plugin-trailer.vue to work and be responsive
- [ ]  The plugin should be disable by default in the glances.conf file (for the moment, on the dev branch, il let it enable by default)

#### Resume

* Bug fix: no
* New feature: yes
* Fixed tickets:

### DIFF
--- a/glances/plugins/tailer/__init__.py
+++ b/glances/plugins/tailer/__init__.py
@@ -1,3 +1,4 @@
+# -*- coding: utf-8 -*-
 #
 # This file is part of Glances.
 #
@@ -55,7 +56,6 @@ items_history_list = [
 # -----------------------------------------------------------------------------
 # Plugin class
 # -----------------------------------------------------------------------------
-
 
 class PluginModel(GlancesPluginModel):
     """Tailer plugin main class.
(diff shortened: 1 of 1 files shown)


nicolargo:
Hi @benjimons 

for Windows OS i think (hope) that the only think to do is take into account the carrier return. It is '\n' on Linux but '\r\n' on Windows.

benjimons:
@nicolargo maybe that should do it for the windows compatibility?

PR #3106, opened 2025-02-17
default_config_dir: Fix config path to include glances/ directory

almkuznetsov:
#### Description
Non-venv default config path was `/usr/share/doc/glances.conf` instead of `/usr/share/doc/glances/glances.conf` since a55970a.
Fix it and place the venv path to the end of the paths list, so it will have priority over the system default.

By the way, I'm not sure if we need to keep system default config path in venv, then we can delete `get_base_prefix_compat()` and `in_virtualenv()` functions, since they only check for sys.base_prefix != sys.prefix, and sys.prefix will point to venv when inside it and to `/usr` when outside

Also fixed some typos in docs on the way

#### Resume

* Bug fix: yes
* New feature: no
* Fixed tickets: no

### DIFF
--- a/docs/config.rst
+++ b/docs/config.rst
@@ -17,9 +17,9 @@ Location
 You can place your ``glances.conf`` file in the following locations:
 
 ==================== =============================================================
-``Linux``, ``SunOS`` ~/.config/glances/, /etc/glances/, /usr/share/docs/glances/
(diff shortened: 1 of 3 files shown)


nicolargo:
Thanks for the PR @almkuznetsov 

Merged into develop.

Will be release in Glances 4.3.1.

PR #3118, opened 2025-02-25
Replace `docker` by `containers` in `glances-grafana-flux.json`

jo-chemla:
#### Description

Typo probably introduced following a naming convention change that was not repercuted on that grafana fluxQL influxdb2 dashboard JSON

#### Resume

* Bug fix: yes
* New feature: no


jo-chemla:
By the way, quick question: I deployed, on a dozen NAS via portainer edge agents, 2 glances instances
 - one for live-monitoring via the server (aggregated by a meta glances server via serverlist) 
 - and one for historical data exported to influxdb. 
 
On the grafana dashboard, all hostnames look like uuid rather than human-readable forms, probably because the glances containers - and probably the portainer agent containers themselves - have that uuid defined as a hostname. What would be the simplest way to rename that to a human readable form, either in glances or grafana? 

I tried to find the tags associated with the measurement with fluxql, but could not manage to do it via the influxdb or grafana query editors.

nicolargo:
Thanks for the PR @jo-chemla

PR #3160, opened 2025-04-25
Fix matching problem when fs config has "show" value

nguuuquaaa:
#### Description

Both GlancesPluginModel.is_hide and GlancesPluginModel.is_show has very bad case-insentitive regex matching implementation, and fs module has a double check on both mount point and device name that most of the time results in 0 match.

#### Resume

* Bug fix: yes
* New feature: no
* Fixed tickets:

### DIFF
--- a/glances/plugins/fs/__init__.py
+++ b/glances/plugins/fs/__init__.py
@@ -164,7 +164,7 @@ class PluginModel(GlancesPluginModel):
         for fs in fs_stat:
             # Hide the stats if the mount point is in the exclude list
             # It avoids unnecessary call to PsUtil disk_usage
-            if not self.is_display(fs.mountpoint):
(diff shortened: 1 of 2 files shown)


nicolargo:
Hi @nguuuquaaa 

thanks for the PR but unfortunately it did not work as expected.

For example on my system with the current develop branch, the side bar is the following:

image

with your branch:

image

My glances.conf file is the following: https://pastebin.com/raw/kUQz0aq7

nguuuquaaa:
@nicolargo You are right, the check is wrong when show is empty. It should be fixed now.

nicolargo:
Last version works fine :)

Thanks for the PR @nguuuquaaa 

Merged into develop branch.

Will be release on next milestone.

PR #3161, opened 2025-04-26
plugins: connections: refactor

ariel-anieli:
#### Description

Extracted `update_with_net_connections_method` and `update_with_nf_conntrack_method` from `update`.

After:
image
Before:
image

#### Resume

Part-of #2801, refactor of [`update`](https://scrutinizer-ci.com/g/nicolargo/glances/code-structure/develop/py-function/glances.plugins.connections.PluginModel.update)

PR #3162, opened 2025-04-27
Revert "plugins: connections: refactor"

nicolargo:
Reverts nicolargo/glances#3161

Because it breaks the unittest on MacOS: https://github.com/nicolargo/glances/actions/runs/14692345739/job/41229404619#logs


ariel-anieli:
I'll have another take on this, @nicolargo; thanks for the heads up.

PR #3249, opened 2025-07-20
Update docker-compose.yml

korn3r:
#### Description
added commented out by default mount point for /etc/os-release. without it Glances container shows distibution information from container itself and not host system.

* Bug fix: no
* New feature: no
* Fixed tickets: none


nicolargo:
Thanks @korn3r  !

PR #3253, opened 2025-07-23
remove `.keys()` from loops over dicts

Boris-Okassa:
#### Description

By default dict loops use keys.

#### Resume

* Bug fix: no
* New feature: no
* Fixed tickets:


nicolargo:
Thanks for the PR @Boris-Okassa

Issue #3166, opened 2025-04-30
Metrics sent by Prometheus exporter are hard to use


ZbigniewRA:
**Feature request**
Metrics sent by Prometheus exporter are hard to use.
For example for mount points `/`, `/var/tmp` and `/var-tmp` it will generate metrics like:
- `glances_fs___percent`
- `glances_fs__var_tmp_percent`
- `glances_fs__var_tmp_percent`

And for GPUs it will generate metrics like:
- `glances_gpu_nvidia0_proc`
- `glances_gpu_nvidia1_proc`
- `glances_gpu_nvidia2_proc`

Similarly for other metrics, like network interfaces, or block devices.

Those metrics are hard to work with:
- It is hard to gather all `fs_percent` or `gpu_proc` metrics for all mount points/gpus.
- It is even harder to figure out the mount point from the metric.

**Describe the solution you'd like**
Ideally it should generate metrics like this:
- `glances_fs_percent (label: mount_point="/")`
- `glances_fs_percent (label: mount_point="/var/tmp")`
- `glances_fs_percent (label: mount_point="/var-tmp")`
- `glances_gpu_proc (label: gpu_name="nvidia0")`
- `glances_gpu_proc (label: gpu_name="nvidia1")`
- `glances_gpu_proc (label: gpu_name="nvidia2")`

Alternatively label could be a generic name, like `object`.

**Describe alternatives you've considered**
I did use the original Prometheus exporter, but it was a pain to use, and not everything was even possible to do for me (I wasn't able to get mount point names for example, or do all aggregations I wanted).

I have modified the `glances_prometheus` exporter to do what I described, which is easy, as the `object` is simply the part before dot, and metric is the part after dot in the input from Glances (i.e. `/var/tmp.percent`)

I could make a pull request with changes, but some design discussion would be needed.
For example:
- Should new behaviour be controlled by a config option?
- Is a generic `object` label ok, or should it be configurable per plugin, or maybe per metric? (i.e. `mount_point` for `fs`, `gpu_name` for `gpu`, `device` for `diskio`).
- I needed to ignore thresholds (`_careful/_critical/_warning`), as those are really meta-data, were not useful for me and were increasing the output needlessly. Is such an option useful to add?
- I needed to ignore some metrics, like virtual interfaces, or individual RAID disks. I've added that in my exporter, but maybe a better place is more useful?

**Additional context**
There is no support for external exporters, so I was not able to add it externally. I had to patch Glances code itself.
[...]

PR #3283, opened 2025-09-23
Fix prometheus export format

DeepSpace2:
### **This is a breaking change for anyone already relying on the affected metrics**

Using proper labels for prometheus metrics of specific plugins (amps, containers, diskio, fs, gpu, network, percpu, sensors, wifi) instead of exporting each individual object as it's own metric which is both hard to use and against prometheus best-practices.

For example, before this PR we exported the following FS metrics:

`glances_fs__etc_hostname_free{hostname="HOSTNAME"}` 

This PR changes this metrics to

`glances_fs_free{hostname="HOSTNAME",mnt_point="/etc/hostname"}`

The used labels are the return value of each plugin's `get_key()`. 
At the moment these are:

- amps: name
- containers: name
- diskio: disk_name
- fs: mnt_point
- gpu: gpu_id
- network: interface_name
- percpu: cpu_number
- sensors: label
- wifi: ssid

#### Resume

* Bug fix: yes
* New feature: no
* Fixed tickets: #3166

### DIFF
--- a/glances/exports/export.py
+++ b/glances/exports/export.py
@@ -225,7 +225,7 @@ class GlancesExport:
 
     def is_excluded(self, field):
         """Return true if the field is excluded."""
-        return hasattr(self, 'exclude_fields') and any(re.fullmatch(i, field, re.I) for i in self.exclude_fields)
+        return any(re.fullmatch(i, field, re.I) for i in (getattr(self, 'exclude_fields') or ()))
 
     def plugins_to_export(self, stats):
         """Return the list of plugins to export.
--- a/glances/exports/export.py
+++ b/glances/exports/export.py
@@ -152,7 +152,7 @@ class GlancesExport:
         d_tags = {}
         if tags:
             try:
-                d_tags = dict([x.split(":") for x in tags.split(",")])
+                d_tags = dict(x.split(":", 1) for x in tags.split(","))
             except ValueError:
                 # one of the 'key:value' pairs was missing
                 logger.info("Invalid tags passed: %s", tags)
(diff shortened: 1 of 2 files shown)


DeepSpace2:
After some thought this does not have to be a braking change right away.

We can keep exporting the old metrics with a depreciation message (both in the docs and in the metrics' documentation so it is visible within prometheus and grafana) in addition to the new metrics until we reach version 5. This way we don't break semantic versioning.

Regarding the "needs reworking" tag. If you don't mind I can rework on it, just let me know which part is not good enough :)

nicolargo:
I have made a review and a comment here: https://github.com/nicolargo/glances/pull/3283/files/afdd09d6482dc9ddbaedaf572299cb69eb607486

image

Regarding the breaking change it is not relevant. Do not worry about it.

DeepSpace2:
@nicolargo oh, didn't see it because it's still pending :)

I pushed a new commit. The mapping is now being built using a `GlancesStats` object. 
As the labels used now are a bit different than the previously hardcoded values I will updated the PR description for documentation purposes 

**However** The values returned by `get_key` of some plugin classes are wrong. For example, `WifiPlugin.get_key()` returns `ssid` instead of `interface_name` but I think this is out of scope for this PR

nicolargo:
Merged into develop.

"However The values returned by get_key of some plugin classes are wrong": Not really, the key has been choose to be "unique". It's not an human oriented choose but more a programmatic one..

nicolargo:
Thanks for the PR @DeepSpace2

DeepSpace2:
Thanks for merging :)

Regarding the keys. I'm not sure I'm following the uniqueness requirement. We already have non-unique keys across plugins (both `amps` and `containers` plugins use `"name"`). It's a bit weird to have the metric 

`glances_wifi_quality_link{hostname="ubuntu-laptop", instance="192.168.0.101:9091", job="glances", ssid="wlp4s0"}`

with label `ssid` but the value is the interface name and not an actual SSID.

PR #3285, opened 2025-09-24
Fix ZFS cache math. Add ZFS to Quicklook

### LINKED ISSUE #3279: ZFS ARC/cache not reported as cache memory usage
Memory used by ZFS as cache is reported as "used" memory instead of "cached" memory, leading to high memory usage being reported. 

My ZFS is configured to use up to 65GB of RAM (and often uses all of that), but it is cache only and not truely "used" memory as it will shrink as needed.

Image

`htop` correctly reports this as cache memory (note that yellow in the memory bar is cache):

Image

`btop` also correctly reports this as cache memory:

Image

Note that you can get current ZFS cache size by running: `arcstat -f "size"` (without root). A possible solution would be to mark this amount as cache in glances.

**Expected behavior**
ZFS ARC usage should show as cache in memory reporting and not "used".

**Environement **
 - Operating System (lsb_release -a or OS name/version): Ubuntu 24.04.3 LTS
 - Glances & psutil versions: 4.3.3
 - How do you install Glances: `pipx install glances`

bkanuka:
Fixes #3279

### DIFF
--- a/glances/plugins/mem/__init__.py
+++ b/glances/plugins/mem/__init__.py
@@ -173,6 +173,7 @@ class MemPlugin(GlancesPluginModel):
 
         # Manage ZFS cache (see #3979 for details)
         if self.zfs_enabled:
+            zfs_size = 0
             zfs_shrink = 0
             zfs_cache_stats = zfs_stats()
             # Uncomment the following line to use the test data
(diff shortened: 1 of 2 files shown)


nicolargo:
@bkanuka Please can you make the PR on the develop branch ?

PR #3289, opened 2025-09-25
Issue #3279 zfs cache and memory math issues

bkanuka:
See issue #3279  for details

Modifies value of `free` memory to be memory completely unused by the system (zero'd).


bkanuka:
Okay I rebased on develop. This should be good to go. @nicolargo

Issue #3290, opened 2025-09-26
Glances hangs if network device is no available


Tiemichael:
**Check the bug**
nothing similar

**Describe the bug**
I am running the most recent version of Glances on a NanoPC-T6 (ARM, RK3588) which runs Ubuntu 24.04, installed via PIP.
After enabling NFS in /etc/grances/glances.conf  [fs] 'allow=nfs', I am able to monitor a network device monted on the machine (mainly to check free space]
While the nfs-drive is mounted, everything works fine. I see the GUI (device:61208) updating every 2-3 seconds.
Once the nfs-drive is unmounted, the GUI hangs, and it's only updated every 20-30 seconds. Accessing the GUI is not possible while Glances is stalled.
This causes issues with the connected HomeAssistant.

**Expected behavior**
GUI is updated every 2-3 seconds

**Screenshots**

**Environement (please complete the following information)**
 - Operating System: 
      Distributor ID: Ubuntu
      Description:    Ubuntu 24.04.3 LTS
      Release:        24.04
     Codename:       noble
 - Glances & psutil versions: 
     Glances version:        4.3.3
     Glances API version:    4
     PsUtil version:         7.1.0
 - How do you install Glances: Pypi package
 - Glances test:  see attachment 

[glances.txt](https://github.com/user-attachments/files/22555183/glances.txt)

ntext**
 Glances is started via Systemd:
     ExecStart=/root/.local/share/pipx/venvs/glances/bin/glances -w  --enable-plugin sensors,smart,wifi --disable-plugin network

* the Glances configuration file : defalt, only added under [fs] 'allow=nfs',
* the Glances log file:  nothing relevant:
   INFO -- Start Glances 4.3.3
   INFO -- CPython 3.12.3 (/root/.local/share/pipx/venvs/glances/bin/python) and psutil 7.1.0 detected
   INFO -- Read configuration file '/etc/glances/glances.conf'
   INFO -- Start GlancesWebServer mode
   INFO -- Glances Web User Interface started on http://0.0.0.0:61208/
   INFO -- Glances RESTful API Server started on http://0.0.0.0:61208/api/4
   INFO -- Autodiscover is enabled with service name michael-desktop
   INFO -- Close the Web server
   INFO -- Glances stopped gracefully

nicolargo:
Hi @Tiemichael 

thanks for the issue report.

[...]

PR #3291, opened 2025-09-28
Fix: Glances stalling on broken NFS connections

RazCrimson:
Closes: #3290


nicolargo:
Thanks al lot @RazCrimson !

PR #3339, opened 2025-11-09
support for Python from 3.9 to 3.14

plamerdimak:
Good morning,

I would like to discover the git workflow in open-source projects.

Thanks for your understanding,

#### Description

Python 3.9 is end of life, and Python 3.14 is released.

Link: https://devguide.python.org/versions/

#### Resume

* Bug fix: no
* New feature: no
* Fixed tickets:

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -16,7 +16,7 @@ jobs:
       - name: Set up Python
         uses: actions/setup-python@v6
         with:
-          python-version: "3.13"
+          python-version: "3.14"
       - name: Install pypa/build
         run: >-
           python3 -m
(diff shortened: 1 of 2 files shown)


nicolargo:
Hi @plamerdimak thanks for the contribution proposal.

For the support of Python 3.9, i prefer to wait a little bit (end of 2025) before removing the test and support.
Tox is not the only file to change, there is also file in the .github/ folder. 

Please feel free to make an update of the current with the following change:
- support for Python from 3.9 to 3.14

Nicolas

plamerdimak:
Thank you for the feedback, I improved on it.

PR #3376, opened 2025-12-13
Get amdgpu name from `amdgpu.ids`

okulev:
#### Description

Instead of fixed string "AMD GPU", provide gpu name by looking up `/usr/share/libdrm/amdgpu.ids`.

The gpu list is maintained in https://cgit.freedesktop.org/drm/libdrm/tree/data/amdgpu.ids . Installation location is the same on major base distributions:

* https://packages.debian.org/sid/all/libdrm-common/filelist
* https://archlinux.org/packages/extra/x86_64/libdrm/files/
* https://packages.fedoraproject.org/pkgs/libdrm/libdrm/fedora-rawhide.html#files

Since minimum Python version is 3.8, I used walrus operator `:=`. Also extracted reading device files into a function. For performance, I added `@functools.cache` decorator to name lookup.

AMD prefix and Graphics suffix are removed to better fit title space. This also aligns with nVidia gpu names reported by `nvmlDeviceGetName`.

Tested with my 8060S.

#### Resume

* Bug fix: no
* New feature: yes


nicolargo:
Thanks for the PR @okulev  !

Will be release in target milestone.

PR #3377, opened 2025-12-14
Include GTT to mem% for integrated AMD GPUs

okulev:
#### Description

On integrated GPU like my Radeon 8060S, it is usual to set very small VRAM in BIOS and leave larger GPU allocations to GTT. In my case, VRAM is 512MB and GTT is 96G. In such cases glances mem% is always almost 100% but there is a lot of available free memory for the iGPU.

The PR detects AMD integrated GPU by looking for presence of Northbridge voltage, as described in https://docs.kernel.org/gpu/amdgpu/thermal.html . I'm not able to test it on discrete GPU, only on my integrated GPU.

There are many AMD (and Intel) systems with integrated GPU and this patch would report much more realistic gpu mem% by always calculating vram+gtt, not only vram.

I took the opportunity to simplify with `glob.glob` temperature and device list traversal of directories.

The idea for this approach to detection integrated GPU is from https://github.com/fastfetch-cli/fastfetch/blob/2.11.0/src/detection/gpu/gpu_linux.c#L60

#### Resume

* Bug fix: no
* New feature: yes

### DIFF
--- a/glances/plugins/gpu/cards/amd.py
+++ b/glances/plugins/gpu/cards/amd.py
@@ -43,11 +43,10 @@ import functools
 import glob
 import os
 import re
-from typing import Optional
+from typing import List, Optional
 
 DRM_ROOT_FOLDER: str = '/sys/class/drm'
-CARD_REGEX: str = r"^card\d$"
-DEVICE_FOLDER: str = 'device'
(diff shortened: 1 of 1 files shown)


nicolargo:
Merged into develop !

Thanks @okulev

PR #3379, opened 2025-12-16
Fix quicklook in case psutil.cpu_freq().max=0.0

okulev:
#### Description

Inside WSL2, quicklook reports CPU frequencies like `2.00/0.00GHz`.

[`psutil.cpu_freq()`](https://psutil.readthedocs.io/en/latest/#psutil.cpu_freq) may return 0.0 for `max` if not available, which is the case inside WSL2. The PR changes quicklook to show only the current freq in this case.

As a separate issue not addressed by the PR, current freq doesn't vary inside (my) WSL2 and is constantly cpu base frequency.

#### Resume

* Bug fix: yes
* New feature: no


nicolargo:
Thanks for the PR @okulev  !

PR #3381, opened 2025-12-19
Update license specification to SPDX format

ruben-arts:
Thanks for the tool!

#### Description

In an attempt to get better support for the SPDX expression in build systems, I would like to update your license specifier to an SPDX-accepted expression.

There are more license expressions to choose from, but my understanding was that this is the one you wanted to use. 
More can be found here: https://spdx.org/licenses/

And the [latest `pyproject.toml` format](https://packaging.python.org/en/latest/specifications/pyproject-toml/#license) advises the use of the direct string format.

#### Resume

* Bug fix: no
* New feature: no
* Fixed tickets: none

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -8,7 +8,6 @@ classifiers = [
   "Intended Audience :: Developers",
   "Intended Audience :: End Users/Desktop",
   "Intended Audience :: System Administrators",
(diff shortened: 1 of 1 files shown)


nicolargo:
Thanks for the PR @ruben-arts !

PR #3411, opened 2026-01-22
Add CPU core number field to processlist

ffleischer:
#### Description

Adds a new `cpu_num` field to the processlist that displays the CPU core number (0-based) where each process is currently executing.

#### Resume

* Bug fix: no
* New feature: yes
* Fixed tickets: 

#### Motivation
This feature helps users identify which CPU core is running each process, useful for:
- Monitoring process CPU affinity behavior
- Identifying CPU hotspots
- Debugging performance issues in multi-core systems

#### Changes
- Added cpu_num capability detection in glances/processes.py
- Implemented CPU column display in processlist plugin
- Added unit tests for the new field
- Updated documentation (config file, user docs, API docs)

### DIFF
--- a/glances/outputs/glances_curses.py
+++ b/glances/outputs/glances_curses.py
@@ -74,6 +74,7 @@ class _GlancesCurses:
         'M': {'switch': 'reset_minmax_tag'},
         'n': {'switch': 'disable_network'},
         'N': {'switch': 'disable_now'},
+        'o': {'sort_key': 'cpu_num'},
         'p': {'sort_key': 'name'},
         'P': {'switch': 'disable_ports'},
         # 'q' or ESCAPE > Quit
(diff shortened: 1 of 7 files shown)


nicolargo:
Hi @ffleischer 

thanks for the PR.

I will have a deeper look inside the PR this week-end.

At a glance, it miss the implementation in the WebUI (not a big deal, if you are not familiar with VueJS i can manage this on my side).

Nicolas

ffleischer:
Hi @nicolargo,

thanks for the fast response.

To be fair I actually don't have experience with VueJS.

But based on the current implementation it looked like not much effort. So I tried 😊

nicolargo:
Merged into develop.

The only change i made is to disable the feature by default from the default configuration file. 

    disable_stats=cpu_num

comment this line to enable the information in both TUI and WebUI.

Thanks a lot for this good PR @ffleischer

PR #3427, opened 2026-02-09
Add libvirt client into docker.

lubomir-moric:
Description
This Pull Request adds the necessary libvirt client packages to the Glances Docker images (both Alpine and Ubuntu flavors) and provides the commented-out volume mapping in the docker-compose.yml for users who need to monitor virtual machines.

Monitoring libvirt from within a container requires the client library to be present inside the image to communicate with the host's libvirt socket.

Resume
Bug fix: yes
New feature: no
Fixed tickets: #3436

### DIFF
--- a/docker-files/alpine.Dockerfile
+++ b/docker-files/alpine.Dockerfile
@@ -28,8 +28,7 @@ RUN apk add --no-cache \
   wireless-tools \
   smartmontools \
   iputils \
-  tzdata \
-  libvirt-client
+  tzdata
 
 ##############################################################################
 # BUILD Stages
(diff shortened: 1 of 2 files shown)


nicolargo:
Hi @lubomir-moric 

I want to keep minimal image "minimal". With the current implementation proposed in this PR, images sizes increase from 232 to 293MB for the Alpine.

Can you please propose a PR that only modify dependencies for the "full" images ?

Thanks !

lubomir-moric:
> Hi @lubomir-moric
> 
> I want to keep minimal image "minimal". With the current implementation proposed in this PR, images sizes increase from 232 to 293MB for the Alpine.
> 
> Can you please propose a PR that only modify dependencies for the "full" images ?
> 
> Thanks !

Makes absolute sense, done.

nicolargo:
Thanks for the PR and the change !

Merged into develop.

PR #3454, opened 2026-02-24
Improve JSON serializer with comprehensive normalization and tests

Julietmgbole:
#### Summary
This PR improves the JSON output layer by introducing a robust serializer that ensures consistent and valid JSON across plugins.

### Key Improvements
- Always produces valid JSON output
- Handles nested data, bytes, datetime, and edge cases
- Ensures error isolation (one plugin failure does not break output)
- Adds configurable error reporting
- Optional metadata inclusion
- Fully backward compatible output format

### Tests
- Added 28 comprehensive unit tests
- All tests passing locally

* Bug fix: No
* New feature: Yes
* Fixed tickets: None

### DIFF
--- a/glances/outputs/glances_stdout_json.py
+++ b/glances/outputs/glances_stdout_json.py
@@ -6,54 +6,179 @@
 # SPDX-License-Identifier: LGPL-3.0-only
 #
 
-"""Stdout interface class."""
+"""Stdout JSON interface class."""
 
 import time
+from typing import Any, List, Optional
 
 from glances.globals import printandflush
+from glances.logger import logger
+from glances.outputs.glances_json_serializer import GlancesJSONSerializer
 
 
 class GlancesStdoutJson:
     """This class manages the Stdout JSON display."""
 
-    def __init__(self, config=None, args=None):
-        # Init
+    DEFAULT_DURATION: int = 3
+    FALLBACK_ERROR_JSON: str = '{"error": "Failed to serialize stats"}'
+
+    def __init__(self, config: Optional[Any] = None, args: Optional[Any] = None):
         self.config = config
         self.args = args
+        self._serializer: Optional[GlancesJSONSerializer] = None
+        self._plugins_list: List[str] = []
 
-        # Build the list of plugin to display
-        self.plugins_list = self.build_list()
+        self._init_plugins_list()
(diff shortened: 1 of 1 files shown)


nicolargo:
Hi @Julietmgbole, and welcome to the Glances project — thank you for your first contribution, it's always appreciated ! 

I've reviewed your PR and the code itself looks clean and well-structured, so great job on that !

However, I have one blocking concern before I can merge this: the new class isn't referenced anywhere in the existing codebase, which means it would be introduced as dead code. For a project used by thousands of users, I need to keep the codebase as lean and navigable as possible, so I can't merge code that isn't actively used.

Could you help me understand the intent ?

- Was this meant as a preparatory step, with a follow-up PR that integrates the class into the rest of the code ?
- Or was the integration accidentally left out of this PR ?

If you have a follow-up planned, I'd love to see the full picture — either as a draft PR or a quick description here — so I can review the design as a whole. If the integration was accidentally omitted, feel free to update this PR to include it.

Julietmgbole:
@nicolargo Thanks for reviewing and for the kind words! 
You're right: the new JSON serializer isn't wired into the runtime path yet. My intent was to make the serializer + tests first, then integrate it into the existing JSON output flow in a follow-up step. 

To avoid adding dead code, I can update this PR to hook it into the current JSON output (replace the existing json serialization call so the class is actually used), and keep the changes limited to the output layer. 

If you prefer it split, I can instead close this and open a new PR that include both the serializer + the integration. Which do you prefer?

nicolargo:
Thanks for your explanations. Your first approch looks good for me (hook it into the current JSon serializer).

Julietmgbole:
Great, thanks! I'll update the PR to integrate it into the current JSON serializer and push the changes shortly.

Julietmgbole:
Wired the new JSON serializer into GlancesStdoutJson,update so that the output is built from structured data and serialized via GlancesJSONSerializer instead of manual string building. 

Verified locally with:
``python -m pytest -q tests/test_json_serializer.py``

(28 passed)

nicolargo:
Thanks @Julietmgbole, merged into develop. 

I will try to make some extra tests this weekend with the GlancesStdoutJson.

After that, we can process with the full implementation.

PR #3462, opened 2026-02-28
Enhance plugins to compute min/max/mean stats since Glances startup

sdoshi2061:
Enhance the existing memory plugin to track runtime minimum and maximum memory usage percentage since Glances startup and expose these values through the existing /api/4/mem endpoint.

#### Description
1. Memory Plugin:
   - Introduce two new attributes: min_percent and max_percent.
   - Initialize both values using the first collected memory percentage.
   - Update them on each refresh cycle:
       min_percent = min(current, min_percent)
       max_percent = max(current, max_percent)
   - Ensure no impact on existing calculations.

2. REST API (v4):
   - Extend the existing /api/4/mem response.
   - Add two new fields:
       "min_percent"
       "max_percent"
   - Keep all existing fields unchanged.
   - Maintain full backward compatibility.

3. Testing:
   - Add unit tests verifying:
       - Proper initialization of min/max
       - Correct update behavior across multiple refresh cycles
       - API response includes new fields
   - Ensure all existing tests pass.

4. Documentation:
   - Update relevant API documentation to describe the new fields.
   
For any questions concerning installation or use, please open a discussion (https://github.com/nicolargo/glances/discussions), not an issue.

#### Resume

* Bug fix: yes/no - No
* New feature: yes/no - Yes
* Fixed tickets: comma-separated list of tickets fixed by the PR, if any - No

### DIFF
--- a/docs/aoa/memory.rst
+++ b/docs/aoa/memory.rst
@@ -40,9 +40,11 @@ Additional stats available in through the API:
   never moved to disk.
 - **shared**: (BSD): memory that may be simultaneously accessed by multiple
   processes.
-- **min_percent**: the minimum memory usage percentage observed since
+- **percent_min**: the minimum memory usage percentage observed since
   Glances startup.
-- **max_percent**: the maximum memory usage percentage observed since
+- **percent_max**: the maximum memory usage percentage observed since
+  Glances startup.
+- **percent_mean**: the mean memory usage percentage observed since
   Glances startup.
 
 It is possible to display the available memory instead of the used memory
--- a/glances/plugins/mem/__init__.py
+++ b/glances/plugins/mem/__init__.py
@@ -28,14 +28,7 @@ and it is supposed to be used to monitor actual memory usage in a cross platform
     'percent': {
         'description': 'The percentage usage calculated as (total - available) / total * 100.',
         'unit': 'percent',
-    },
-    'min_percent': {
-        'description': 'Minimum memory usage percentage observed since Glances startup.',
-        'unit': 'percent',
-    },
-    'max_percent': {
-        'description': 'Maximum memory usage percentage observed since Glances startup.',
-        'unit': 'percent',
+        'mmm': True
     },
     'used': {
         'description': 'Memory used, calculated differently depending on the platform and \
@@ -139,10 +132,6 @@ class MemPlugin(GlancesPluginModel):
 
         # ZFS
         self.zfs_enabled = zfs_enable()
-
-        # Runtime min/max memory usage percentage tracking (since Glances startup)
-        self._min_percent = None
-        self._max_percent = None
 
         # We want to display the stat in the curse interface
         self.display_curse = True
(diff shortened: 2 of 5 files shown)


nicolargo:
Hi @sdoshi2061 

first of all thanks for the enhancement proposal, i think that it will be a nice feature for the Glances users.

Nevertheless, i am not completely agree with the current implementation limited to the memory. 

My alternative proposal is the following:

1) add a new 'mmm' (for Min/Max/Mean) and the fields_description for targeted sensors

Ex:

```
    ...
    'percent': {
        'description': 'The percentage usage calculated as (total - available) / total * 100.',
        'unit': 'percent',
        'mmm': True
    },
    ...
```
 
2) when this tag is set and the Value is True then generate automatically the following fields:

-  percent_min
- percent_max
- percent_mean

The computation of min, max, mean should be done in the mother class: glances/plugins/plugin/model.py.

This implementation will allow us to extend the mmm feature to others plugins.

What do you think ?

sdoshi2061:
Hi @nicolargo 

Thank you for the detailed feedback and for the alternative proposal. I completely agree that implementing the mmm logic at the mother class level is a much cleaner and more scalable approach.

Moving the min/max/mean computation to glances/plugins/plugin/model.py and introducing the mmm flag in fields_description makes the feature reusable across plugins, which is definitely better than limiting it to memory only.

I will refactor the current implementation accordingly :
- Add the mmm attribute support in fields_description
- Implement automatic generation of percent_min, percent_max, and percent_mean fields when this tag is set and the Value is True
- Move the computation logic to the base plugin model.

Before proceeding, would you prefer that I update this PR directly with the refactored implementation, or should I go through the Anthropic feedback process and submit it as a separate task iteration?

nicolargo:
Update of this PR will be fine. Please also make sure that Unit test are added.

sdoshi2061:
Hi @nicolargo 

Updated the PR based on your feedback.
 - The mmm logic is now implemented in the base plugin model and is reusable across plugins.
 - Min/Max/Mean fields are automatically generated when the mmm flag is enabled.
 - Unit test has been added.

MMM true
with_mmm

MMM false
without_mmm

Looking forward to your feedback.

sdoshi2061:
Hi @nicolargo 

I just wanted to follow up on the updated PR.

As discussed, I’ve refactored the implementation so the MMM logic now lives in the base plugin model, making it reusable across all plugins.

mmm: true → automatically generates min/max/mean fields (with_mmm)

mmm: false → keeps original behavior (without_mmm)

Added unit tests to cover both scenarios

Please let me know if you’d like any further refinements or structural changes.
Looking forward to your feedback. 😊

nicolargo:
Yes latest PR is ok for me. I will have a deeper look in it this weekend before merging to develop branch.

nicolargo:
Merged int develop and min/max/mean also applied on CPU (total) and LOAD (min1) !

Thanks for the contrib @sdoshi2061

Issue #3469, opened 2026-03-04
System display error on "little" terminal.


nicolargo:
On a terminal, if the system name can not be displayed then the host and the IP are merged:

<img width="319" height="74" alt="Image" src="https://github.com/user-attachments/assets/d0144a52-8181-465b-ae64-08c90cfdcc6c" />

psyberck:
Hey @nicolargo, can I be assigned to work on this issue as my first contribution to this project? I can start working on it tomorrow. Thanks!

nicolargo:
Hi @ZihaoXue1995 ! Welcome [on-board](https://github.com/nicolargo/glances/wiki/How-to-contribute-to-Glances-%3F) :)

sepastian:
Hi @nicolargo, this doesn't seem to be a problem on my system, at least when running version 4.5.1.

Only in full width, hostname, IP address etc. are shown. When setting the window width to half size, only the hostname (left column) and up time (right column) are retained. When making it even smaller, to quarter width, parts of the hostname will be hidden, up time stays visible.

In my opinion, this works just fine.

What is the expected behavior here? Or does it look differently in version 5.4.2?

Full width:

<img width="1920" height="198" alt="Image" src="https://github.com/user-attachments/assets/65784eb5-e9dc-4474-a1b6-536f7d8ade70" />

Half width:

<img width="1918" height="228" alt="Image" src="https://github.com/user-attachments/assets/1b412e1b-e511-4bd6-a80d-c6f4803bc711" />

Quarter width:

<img width="1920" height="281" alt="Image" src="https://github.com/user-attachments/assets/fa1eb2a3-2feb-41da-94bc-81674b102f73" />

nicolargo:
Hi @sepastian 

try to reduce step by step the terminal, on my configuration it happens here:

<img width="873" height="211" alt="Image" src="https://github.com/user-attachments/assets/b162bec2-59ae-4994-8160-08e641d438b2" />

Perhaps a static analysis of the code can help.

https://github.com/nicolargo/glances/blob/d2b172e3fb5f53dbf1e38d22c42713c9907fb2cf/glances/outputs/glances_curses.py#L692
https://github.com/nicolargo/glances/blob/d2b172e3fb5f53dbf1e38d22c42713c9907fb2cf/glances/plugins/ip/__init__.py#L166

passionworkeer:
[...]

PR #3473, opened 2026-03-11
refactor(glances_curses): add space between system info and IP when necessary

psyberck:
#### Description

Add one extra space between host-name and IP address when needed.

##### Root cause
https://github.com/nicolargo/glances/blob/d2b172e3fb5f53dbf1e38d22c42713c9907fb2cf/glances/outputs/glances_curses.py#L700
https://github.com/nicolargo/glances/blob/d2b172e3fb5f53dbf1e38d22c42713c9907fb2cf/glances/outputs/glances_curses.py#L1097
Bare host-name does not end with a space so the IP column `self.next_column` will render right away without any offset space in between.

##### Solution
- The easiest solution is to set `self.space_between_column=1` initially, but it will result in a UI change (a new space added) for users who have optional system info to display.
- To make the UI consistent, it only adds a space when there isn't one between system info and IP info.

##### Results
Screenshot from 2026-03-11 13-00-25
Screenshot from 2026-03-11 13-18-12

#### Resume

* Bug fix: yes
* New feature: no
* Fixed tickets: #3469

### DIFF
--- a/glances/outputs/glances_curses.py
+++ b/glances/outputs/glances_curses.py
@@ -697,12 +697,20 @@ class _GlancesCurses:
         """
         # First line
         self.new_line()
-        self.space_between_column = 0
         l_uptime = 1
         for i in ['system', 'ip', 'uptime']:
             if i in stat_display:
                 l_uptime += self.get_stats_display_width(stat_display[i])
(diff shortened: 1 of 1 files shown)


nicolargo:
Just merged into deveop.

Thanks for the PR @ZihaoXue1995 !

psyberck:
My pleasure, I will see what I can contribute more to this project.

PR #3482, opened 2026-03-15
Add LXD/LXC container engine support

### LINKED ISSUE #3480: Support for LXC/LXD containers
Glances docs mention LXC in passing, but it appears that only `docker` and `podman` containers engines are implemented.

I propose adding `glances/plugins/containers/engines/lxd.py` with per-container stats (CPU, memory, disk IO, network) via the LXD API. Could use the official [pylxd](https://github.com/canonical/pylxd) REST client.

Would you be interested in a PR for this feature?

crishoj:
Closes #3480

Adds LXD as a third container engine alongside Docker and Podman.

- New engine at `glances/plugins/containers/engines/lxd.py` using `pylxd`
- Polls `instance.state()` for CPU, memory, disk IO, and network stats
- In clustered LXD, filters to instances on the local node only
- Extracts port mappings from proxy devices
- Registers automatically when `pylxd` is installed and LXD socket is reachable

### Screenshot

image


glances/plugins/containers/engines/lxd.py
```diff
@@ -0,0 +1,317 @@
+#
+# This file is part of Glances.
+#
+# SPDX-FileCopyrightText: 2024 Nicolas Hennion <nicolas@nicolargo.com>
```
nicolargo: Change it with your name / contact and 2026

glances/plugins/containers/engines/lxd.py
```diff
@@ -0,0 +1,317 @@
...
+        computed = self._compute_activity_stats()
+        self._old_computed_stats = computed
+        self._last_stats_computed_time = time.time()
+        return computed
+
+    @property
+    def time_since_update(self) -> float:
+        return max(1, time.time() - self._last_stats_computed_time)
+
+    def _compute_activity_stats(self) -> dict[str, dict[str, Any]]:
```
nicolargo, line 75: In order to reduce the complexity of this method, it will be nice to split (put CPU, MEM, DISK and NET in sub-functions).

glances/plugins/containers/engines/lxd.py
```diff
@@ -0,0 +1,317 @@
...
+    disable_plugin_lxd = True
+    logger.warning(f"Error loading LXD deps Lib. LXD feature in the Containers plugin is disabled ({e})")
+else:
+    disable_plugin_lxd = False
+
+
+class LxdStatsFetcher:
+    """Fetch stats for a single LXD instance by polling its state."""
+
+    def __init__(self, instance, poll_interval=2):
```
nicolargo, line 32: Instead of the default 2 seconds value for the poll_interval, is it possible to align it with the default Glances refresh interval (args.time) ?

nicolargo (changes requested):
Great work !

Some minor improvements in the code (see review) and did not forget to change the plugins documentation (docs/aoa/containers.rst).

crishoj:
Thanks for the feedback. Should be addressed now.

One open question: I have only tested this with _containers_ — not VMs. The LXD API should report similar stats for VMs, though, so it should just work.

nicolargo:
Thanks, merged into develop !

I will make some test on my side before closing the issue https://github.com/nicolargo/glances/issues/3480

PR #3497, opened 2026-03-24
Use sys.executable in the testsuite

s-t-e-v-e-n-k:
#### Description

Rather than looking for a venv python executable, use the existing sys.executable property to execute the modules required.

#### Resume

* Bug fix: yes (sort of?)
* New feature: no
* Fixed tickets: None


nicolargo:
Thanks for the PR @s-t-e-v-e-n-k !

Merged into develop.

PR #3499, opened 2026-03-25
fix: support single-core Rockchip NPU load parsing and improve device naming

csvke:
## Summary

The Rockchip NPU plugin's `parse_rknpu_load()` only matches the **multi-core** format used by RK3588/RK3576:

```
NPU load: Core0: 45%, Core1: 32%, Core2: 0%,
```

Single-core Rockchip NPUs (e.g. **RV1126B**) output a different format:

```
NPU load:  0%
```

This was never matched by the existing regex, causing `load` to silently return `0` even under active inference workloads.

### 1. Single-core NPU load parsing
Added a fallback regex `r'NPU load:\s*(\d+)%'` after the multi-core pattern check. Multi-core format still takes priority when cores are present.

### 2. NPU device naming
Changed the display name source from `/proc/device-tree/model` (board name, e.g. "Orange Pi 5 Plus") to the NPU's own device-tree compatible string at `/sys/class/devfreq/*/device/of_node/compatible` (e.g. `rockchip,rv1126b-rknpu` → **"Rockchip RV1126B RKNPU"**). Falls back to board model if the compatible string is unavailable.

### 3. Return type fix
Corrected the `parse_rknpu_load()` return type annotation from `list[dict]` to `int | None`.

## Testing

Tested on a **Rockchip RV1126B-P** board with RKNPU driver v0.9.8:

- **Idle**: `NPU load:  0%` → Glances reports `load: 0` ✅
- **Under inference** (YOLOv5s 640×640): `NPU load: 48%` → Glances reports `load: 48-53` ✅
- **Multi-core backward compatibility**: Existing `Core0: 45%, Core1: 32%` format still works ✅
- **Device name**: Shows `Rockchip RV1126B RKNPU` instead of board model ✅

## Affected devices

- **RV1126B** (confirmed, single-core NPU)
- Likely other single-core Rockchip NPUs (RV1109, RV1126, etc.)
- **RK3588/RK3576** remain unaffected (multi-core path unchanged)

### DIFF
--- a/glances/plugins/npu/cards/rockchip.py
+++ b/glances/plugins/npu/cards/rockchip.py
@@ -78,8 +78,7 @@ class RockchipNPU:
         stats = NPUStats()
 
         stats.npu_id = 'rockship_1'
-        model = self._read_file(os.path.join(self.device_folder, 'model'))
-        stats.name = model if model else "Rockchip NPU"
+        stats.name = self._get_npu_name()
         stats.freq_current = self._read_file(os.path.join(self.freq_folder, 'cur_freq'), as_int=True)
         stats.freq_max = self._read_file(os.path.join(self.freq_folder, 'max_freq'), as_int=True)
         stats.freq = int(stats.freq_current / stats.freq_max * 100)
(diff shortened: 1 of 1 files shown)


csvke:
Screenshot of running inferencing with NPU with Glances showing the correct load % as per 
`cat /sys/kernel/debug/rknpu/load ` running `watch -n 0.1 -d cat /sys/kernel/debug/rknpu/load` side by side in terminal.

Screenshot 2026-03-25 at 7 50 59 PM

nicolargo:
Merged into develop branch.

Thanks for the contrib @csvke !

csvke:
You may want to find someone with a multi-core NPU (like RK3588) to test if my logic has no bugs but I am fairly sure it’s ok. Just good practice to test it too.  @nicolargo

Issue #1711, opened 2020-08-21
Add meter for CPU and MEM of GPU in the Quicklook plugin


orena1:
#### Description
I really like the monitor, it is great!
I would find it very efficient if glances will also have a meter for the cpu and mem of the GPU , something like this:

![image](https://user-images.githubusercontent.com/8983713/90943699-8257c580-e423-11ea-8217-7412f57747d3.png)

nicolargo:
Done by @cswaltzinger in https://github.com/nicolargo/glances/pull/3543

PR #3543, opened 2026-04-30
Issue 1711 : Add meter for CPU and MEM of GPU in the Quicklook plugin

cswaltzinger:
#### Description

Average GPU processor and memory meter is added to the quicklook plugin.  

#### Resume
* Bug fix: no
* New feature: yes
* Fixed tickets:  #1711

### DIFF
--- a/.github/workflows/cyber.yml
+++ b/.github/workflows/cyber.yml
@@ -20,7 +20,7 @@ jobs:
         uses: actions/checkout@de0fac2e4500dabe0009e67214ff5f5447ce83dd # v6.0.2
 
       - name: Run Trivy vulnerability scanner in repo mode
-        uses: aquasecurity/trivy-action@57a97c7e7821a5776cebc9bb87c984fa69cba8f1 # master
+        uses: aquasecurity/trivy-action@ed142fd0673e97e23eac54620cfb913e5ce36c25 # master
         with:
           scan-type: 'fs'
           ignore-unfixed: true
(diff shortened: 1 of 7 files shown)


nicolargo:
Hi @cswaltzinger 

First of all thanks for the PR !

I just have a look on the code and as previously done for theCPU (by the use of  "cpu_percent") i will prefer to have a shared variable between GPU and QUICKLOOK in order to not update the GPU twice.  Update should only be done in the target plugin (GPU) and QUICKLOOK should "only" get the result.

Thanks !

cswaltzinger:
Hello @nicolargo,

That makes complete sense.  I have adjusted the code for the GPU and Quicklook plugin accordingly by adding a glances.gpu_percent module to be the access point for the shared memory.  

Please let me know if there is anything else you would like me to address.  

Best,
cswaltzinger

nicolargo:
Hello @cswaltzinger 

Thanks for the change.

One more thing, implementation as been done for the TUI (Curse interface) but not for the WebUI. I want to keep the two UIs aligned in term of feature.

Can you have a look ?

PS: if you are not familiar with VueJS, i can also have a look on my side. Let me know.

cswaltzinger:
Yes, that makes sense.  However, I ran the webui locally and it seems to already have the GPU status fields in the quicklook plugin (see attached screenshot).   Is this what you were referring to?  if not, would you kindly specify which  modules I should edit? 

Best, 
cswaltzinger

Screenshot 2026-05-04 at 5 13 36 PM

nicolargo:
Ok perfect !

Just merged into develop.

Will be release in next target milestone (Glances 4.5.5).

Thanks for the PR @cswaltzinger  !

PR #3557, opened 2026-05-15
feat: add cpu limit to docker, podman and lxd containers

### LINKED ISSUE #3553: Exposing containers CPU limits
We are currently exposing memory limits for containers but not CPU limits.

This is a bit more tricky as it requires some calculations (as it is more dynamic than memory limits, which for example are returned natively by the Docker API).

I already started to implement this (and have it working) for Docker containers.

I was wondering if this is something you would like to see a PR for. If so, I'd see if I can get it to work for Podman and LXD containers (if it is important to keep them aligned/have the same features).

DeepSpace2:
#### Description

Exposes containers CPU limits.
Closes #3553

#### Resume

* Bug fix: no
* New feature: yes
* Fixed tickets: #3553

### DIFF
--- a/glances/plugins/containers/__init__.py
+++ b/glances/plugins/containers/__init__.py
@@ -52,6 +52,10 @@ fields_description = {
         'description': 'Container CPU consumption',
         'unit': 'percent',
     },
+    'cpu_limit': {
+        'description': 'Container CPU limit',
+        'unit': 'number',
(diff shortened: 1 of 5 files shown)


nicolargo:
Thanks for the enhancement @DeepSpace2 

Merged into develop branch.

PR #3558, opened 2026-05-16
Keep auto_unit within limits, so columns stay aligned

metayan:
#### Description

Occasionally, columns got misaligned, because auto_unit returned too many decimals when the number was slightly below 10 or 100. Actually, when (9.995 <= n < 10) and (99.95 < n < 100).

For example,
`10*2**20-1` returned `10.00M` instead of `10.0M` and
`100*2**20-1` returned `100.0M` instead of `100M`.

Tests added to verify correctness.

#### Resume

* Bug fix: yes


nicolargo:
Thanks for the patch @metayan 

Just merged into develop :)

PR #3580, opened 2026-06-07
feat(gpu): support NVIDIA Jetson (Tegra) integrated GPU via sysfs

gstangel:
* Bug fix: no
* New feature: yes
* 

On Jetson (Tegra) the integrated GPU is enumerated by NVML (nvidia-l4t-nvml) and reports its name (e.g. "Orin (nvgpu)"), but the per-metric NVML queries return NVML_ERROR_NOT_SUPPORTED, so the GPU plugin only ever showed N/A for proc/mem/temperature.

Add a Tegra sysfs backend (glances/plugins/gpu/cards/tegra.py) and wire it into the NVIDIA card as a per-metric fallback: when NVML returns None for a device detected as Tegra (name contains "nvgpu", or the Tegra GPU sysfs node exists), read:

- proc:        /sys/devices/platform/gpu.0/load  (per-mille -> percent)
- temperature: the gpu-thermal /sys/class/thermal zone (milli-C -> C)

Memory stays N/A by design: the Tegra GPU shares system RAM, already reported by the MEM plugin. Scales verified against tegrastats (GR3D_FREQ and gpu@).

Adds unit tests with committed sysfs fixtures, a NEWS.rst entry and a docs note.

Verified on JetPack 6.2.1 (L4T R36.4.7) and JetPack 7.2 (L4T R39.2): identical sysfs node layout and gpu-thermal zone selection on both, with temperature cross-checked against tegrastats on each.


nicolargo:
Thanks for the PR @gstangel !

Merged into develop.

Will be release in Glances 4.5.5.

PR #3613, opened 2026-07-14
Sensors: grab temperatures on Windows from LibreHardwareMonitor

### LINKED ISSUE #3265: Plugin for librehardwaremonitor integration to get glances to monitor temperatures on windows
*** If I mess up the message I apologize for my bad english
*** I'm also sorry for any out of protocol event that may happen, it's my first time posting an issue on github
I'm really excited about this, and hope this ends up being considered a contribution

I needed to monitor some things on my small and humble server running windows 10
Temperatures are not read in windows for glances, so I started looking for a workaround

Glances doesn't read temperatures on windows

I already managed for it to happen, I'm here to share what I got to work
Programming or coding really isn't my thing, but I gave it a try, and with help of course IA, I got it working

The solution is a modification on index.html located in C:\PythonXXX\Lib\site-packages\glances\outputs\static\template

And the creation of a script in C:\Python31\Lib\site-packages\glances\outputs\static\public\

Temps in widget bottom left corner

Image

nachobkcc:
## Description

On Windows, `psutil` does not implement `sensors_temperatures()`, so the sensors plugin shows no temperatures at all. This PR grabs the temperatures from the [LibreHardwareMonitor](https://github.com/LibreHardwareMonitor/LibreHardwareMonitor) embedded web server (`data.json` endpoint) and feeds them into the existing sensors plugin, as discussed in #3265.

No WebUI change needed: the temperatures flow through the standard `temperature_core` model, so they show up in the TUI, the Web UI and `/api/4/sensors` automatically.

## How it works

- New `librehardwaremonitor` sub-sensor following the same pattern as the `hddtemp` one (`glances/plugins/sensors/sensor/glances_librehardwaremonitor.py`).
- Registered only on Windows, and only when the psutil temperature grabber is not available — zero impact on Linux/macOS/BSD.
- The user runs LibreHardwareMonitor with its *Remote Web Server* option enabled (default port 8085); host/port are configurable in a new `[librehardwaremonitor]` section.
- Handles both recent `data.json` payloads (`Type` field) and older releases (fallback on the `°C` unit in the value), plus locale decimal separators (`44,5 °C`).
- Sensor labels are `<hardware> <sensor>` (e.g. `Nuvoton NCT6793D SYSTIN`, `Intel Core i7-7700 CPU Package`), so the existing `[sensors]` hide/show/alias machinery applies to them.
- On connection failure it degrades to an empty list and logs once at debug level (no log flooding, and it recovers if LibreHardwareMonitor is started later).

## Design notes

- The [librehardwaremonitor-api](https://pypi.org/project/librehardwaremonitor-api/) lib suggested in #3265 is async (aiohttp) while the sensors grabbers are sync, and it requires Python >= 3.11 while Glances supports >= 3.10. So this PR uses a stdlib-only fetch (`urllib`) — the parsing needed is ~40 lines and there is no new dependency. Happy to switch to the lib if you prefer.
- Fan speeds are available from the same endpoint and would be a small follow-up (`fan_speed` type).
- While wiring the configuration I noticed that all the sensors sub-plugins get `plugin_name == 'sensors'` (`model.py` builds it from the first component of the module path relative to `glances.plugins`), so for instance the `[hddtemp]` host/port section seems to be silently ignored since the 4.x plugins layout. That is why the new section is read directly from the config object. I can open a separate issue for this if you confirm.

## Tests

- `tests/test_plugin_sensors_librehardwaremonitor.py`: 11 new unit tests (tree parsing, sensor type detection, locale decimals, failure modes) — all passing on Windows (Python 3.13).
- `tests/test_plugin_sensors.py`: 32 passed, 3 skipped (Linux-only), unchanged.
- `ruff check` and `ruff format --check` clean.
- End-to-end on a Windows 10 Pro machine, against a local stub replaying the LibreHardwareMonitor `data.json` format: the sub-sensor registers on Windows and the temperatures show up in the TUI, the Web UI and `/api/4/sensors`.
- **Not yet tested against a live LibreHardwareMonitor web server** — my production box still runs an older LibreHardwareMonitorLib-based bridge and I did not want to install the PawnIO driver on it yet. I will follow up with a real-instance validation (and screenshot) in a comment before this is merged.

Closes #3265

🤖 Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/conf/glances.conf
+++ b/conf/glances.conf
@@ -488,6 +488,15 @@ disable=False
 host=127.0.0.1
 port=7634
 
+[librehardwaremonitor]
+# Windows only: grab temperatures from the LibreHardwareMonitor web server
+# The "Remote Web Server" option should be enabled in LibreHardwareMonitor
+# https://github.com/LibreHardwareMonitor/LibreHardwareMonitor
+disable=False
+# Define LibreHardwareMonitor web server IP and port (default is 127.0.0.1 and 8085 (TCP))
+host=127.0.0.1
+port=8085
+
 [sensors]
 # Documentation: https://glances.readthedocs.io/en/latest/aoa/sensors.html
 disable=False
--- a/docs/aoa/sensors.rst
+++ b/docs/aoa/sensors.rst
@@ -3,7 +3,7 @@
 Sensors
 =======
 
-*Availability: Linux*
+*Availability: Linux, Windows (temperatures, through LibreHardwareMonitor)*
 
 .. image:: ../_static/sensors.png
 
@@ -12,6 +12,21 @@ Glances can display the sensors information using ``psutil``,
 - motherboard and CPU temperatures
 - hard disk temperature
 - battery capacity
+
+On Windows, ``psutil`` does not provide temperatures. Glances can grab
+them from the `LibreHardwareMonitor
+<https://github.com/LibreHardwareMonitor/LibreHardwareMonitor>`_ web
+server instead. Run LibreHardwareMonitor, enable the *Remote Web Server*
+option (default port 8085) and (re)start Glances. The web server address
+can be changed in the configuration file under the
+``[librehardwaremonitor]`` section:
+
+.. code-block:: ini
+
+    [librehardwaremonitor]
+    # Define LibreHardwareMonitor web server IP and port (default is 127.0.0.1 and 8085 (TCP))
+    host=127.0.0.1
+    port=8085
 
 Limit values and sensors alias names can be defined in the configuration
 file under the ``[sensors]`` section.
--- a/glances/plugins/sensors/__init__.py
+++ b/glances/plugins/sensors/__init__.py
@@ -96,20 +96,29 @@ class SensorsPlugin(GlancesPluginModel):
 
         self.sensors_grab_map = {}
 
+        librehardwaremonitor_hdd_plugin = None
         if glances_grab_sensors_cpu_temp.init:
             self.sensors_grab_map[sensors_definition.get('cpu_temp').get('type')] = glances_grab_sensors_cpu_temp
         elif WINDOWS:
             # On Windows, psutil does not provide temperatures (see issue #3265)
-            # Grab them from the LibreHardwareMonitor web server (if available)
+            # Grab them from the LibreHardwareMonitor web server (if available):
+            # one instance for the CPU/motherboard/GPU temperatures and one for
+            # the disks ones (displayed as temperature_hdd, see below)
             start_duration.reset()
             librehardwaremonitor_plugin = LibrehardwaremonitorPlugin(args=args, config=config)
+            librehardwaremonitor_hdd_plugin = LibrehardwaremonitorPlugin(args=args, config=config, storage=True)
             logger.debug(f"LibreHardwareMonitor sensor plugin init duration: {start_duration.get()} seconds")
             self.sensors_grab_map[sensors_definition.get('cpu_temp').get('type')] = librehardwaremonitor_plugin
 
         if glances_grab_sensors_fan_speed.init:
             self.sensors_grab_map[sensors_definition.get('fan_speed').get('type')] = glances_grab_sensors_fan_speed
 
-        self.sensors_grab_map[sensors_definition.get('hdd_temp').get('type')] = hddtemp_plugin
+        if librehardwaremonitor_hdd_plugin is not None:
+            # On Windows, the hddtemp daemon is not available: grab the disks
+            # temperatures from the LibreHardwareMonitor web server too
+            self.sensors_grab_map[sensors_definition.get('hdd_temp').get('type')] = librehardwaremonitor_hdd_plugin
+        else:
+            self.sensors_grab_map[sensors_definition.get('hdd_temp').get('type')] = hddtemp_plugin
         self.sensors_grab_map[sensors_definition.get('battery').get('type')] = batpercent_plugin
 
         # We want to display the stat in the curse interface
--- a/glances/plugins/sensors/__init__.py
+++ b/glances/plugins/sensors/__init__.py
@@ -245,6 +245,17 @@ class SensorsPlugin(GlancesPluginModel):
             elif i['type'] == sensors_definition.get('battery').get('type'):
                 # Battery is in %
                 alert = self.get_alert(current=100 - i['value'], header=i['type'])
+            elif (
+                i['type'] == sensors_definition.get('hdd_temp').get('type')
+                and i.get('critical') is not None
+                and not self.is_limit('critical', stat_name=i['type'] + '_' + i['label'])
+            ):
+                # Disks sensors may carry their own thresholds (e.g. NVMe drives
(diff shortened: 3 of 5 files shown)


nachobkcc:
Tested this branch end-to-end on a real Windows box against a real LibreHardwareMonitor instance. **The core feature works**: temperatures now show up on Windows in the API, the web UI and the TUI, where `master` shows nothing at all.

The real-hardware pass also surfaced four issues (one of them made Glances display a permanent false CRITICAL alert on any machine with an NVMe drive), plus a follow-up on the alert logic. All of them are addressed in the two commits pushed after the initial one — details below.

## Test environment

| | |
|---|---|
| OS | Windows 10 Pro 22H2 (build 19045), x64 |
| **System locale** | **es-AR → decimal separator is `,`** (exercises the locale parsing path) |
| CPU | Intel Core i5-11600K (6C/12T) |
| Motherboard | ASUS TUF GAMING Z590-PLUS (Super I/O: Nuvoton NCT6798D) |
| GPU | NVIDIA GeForce RTX 3060 Ti |
| Storage | WD_BLACK SN770 500GB (NVMe) + 3x SATA SSDs |
| LibreHardwareMonitor | **v0.9.6** (`LibreHardwareMonitor.zip` release), Remote Web Server on `:8085` |
| Python / Glances | 3.12.10 / 4.5.6_dev01 (`pip install -e ".[web]"`) |
| Commits tested | `409e3a1` (initial), `b436024` (retest), `91da4fa` (final) |

Note for anyone reproducing this: on current LHM releases you also need the **PawnIO** driver (LHM prompts to install it on first run). Without it, Intel MSR core temperatures are not exposed at all — WinRing0 is now on the Windows vulnerable-driver blocklist — so you would get an empty or nearly empty tree.

## What works

- **Temperatures are grabbed and exposed.** `GET /api/4/sensors` returns the full set of readings with the right types and `"unit": "C"`.
- **Nested hardware resolves correctly.** The motherboard node nests the Super I/O chip one level deeper than the CPU node (`Sensor > HOST > ASUS TUF Z590-PLUS > Nuvoton NCT6798D > Temperatures > …` vs `Sensor > HOST > CPU > Temperatures > …`) and both depths yield the right hardware name (`Nuvoton NCT6798D`, not the mainboard name).
- **Locale decimal parsing works.** This is a comma-decimal box and LHM serializes culture-formatted strings (`"Value":"43,0 °C"`). Values round-trip exactly: LHM `59,9 °C` → Glances `59.9`. Physically stable sensors (disks, Super I/O, GPU) match the LHM window exactly; CPU cores differ only by refresh skew (`sensors` refreshes every 10 s, LHM every ~1 s).
- Graceful when LHM is not running: empty panel, no traceback, failure logged only on state change.

## Issues found on real hardware, and how they were addressed

**1. NVMe threshold constants were reported as live temperatures → permanent false CRITICAL** (fixed in `b436024`)

For NVMe drives LHM exposes the drive's *configured thresholds* as sensor nodes carrying `"Type":"Temperature"` just like real readings:

```json
{"id":207,"Text":"Warning Temperature","Value":"83,0 °C","SensorId":"/nvme/0/temperature/10","Type":"Temperature"}
{"id":208,"Text":"Critical Temperature","Value":"87,0 °C","SensorId":"/nvme/0/temperature/11","Type":"Temperature"}
```

These are static constants from the NVMe spec page, not measurements, but they ended up in the sensors list as if the drive were at 83/87 °C — rendered red/CRITICAL permanently on an idle, healthy machine, for any user with an NVMe drive. They are now consumed as the `warning`/`critical` fields of the real sensors of the same drive instead of being reported as readings.

**2. "Distance to TjMax" sensors are margins, not temperatures** (fixed in `b436024`)

Intel CPUs expose per-core `Distance to TjMax` nodes, also typed `Temperature`. Higher means *cooler*, so the alert logic would read them backwards, and they doubled the CPU rows. Now excluded.

**3. Everything was typed `temperature_core`, including disks** (fixed in `b436024`)

Disk temperatures were evaluated against the CPU thresholds and `temperature_hdd_*` was unreachable. The grabber now splits on the hardware type (`hdd` icon in the data.json tree) and a second instance routes the disks to the `temperature_hdd` slot on Windows (where the Linux-only hddtemp daemon does not apply — that slot was dead there anyway).

**4. Long labels were indistinguishable in the TUI** (fixed in `b436024`)

`<hardware> <sensor>` labels like `11th Gen Intel Core i5-11600K CPU Core #1` all truncated to the same ~22-char prefix in the TUI. Labels are now `<sensor> (<hardware>)` so the distinctive part survives truncation. (Two identical SATA drive models can still collide after truncation; the `[sensors]` alias mechanism covers that corner.)

**5. Follow-up: the harvested NVMe thresholds were stored but not used by the alert logic** (fixed in `91da4fa`)

After fix 1/3, the retest showed the NVMe internal sensor at 59.9 °C flagged WARNING against the generic `temperature_hdd` limits (45/52/60) even though the drive declares warning at 83 °C. `update_views()` now evaluates `temperature_hdd` sensors that carry their own thresholds against them. Precedence: a per-sensor limit from `glances.conf` still wins, but the device-declared thresholds win over the generic type limits — those are always defined (`set_default_cwc` in `config.py`), so the device values could never apply otherwise. Sensors without their own thresholds (SATA disks, hddtemp on Linux) keep the current behaviour.

## Final retest result

With all fixes in (`91da4fa`), on the same machine: 24 sensor rows (32 minus the 6 TjMax margins and 2 threshold constants), disks reported as `temperature_hdd`, and the NVMe internal sensor that used to be flagged now resolves OK against the drive's own thresholds. Sampled 10 times over 40 s at idle: zero WARNING, zero CRITICAL across all samples.

```json
{"label": "CPU Package (11th Gen Intel Core i5-11600K)", "unit": "C", "value": 62.0,
 "warning": null, "critical": null, "type": "temperature_core", "key": "label"}
{"label": "GPU Hot Spot (NVIDIA GeForce RTX 3060 Ti)", "unit": "C", "value": 61.0,
 "warning": null, "critical": null, "type": "temperature_core", "key": "label"}
{"label": "Composite Temperature (WD_BLACK SN770 500GB)", "unit": "C", "value": 42.0,
 "warning": 83, "critical": 87, "type": "temperature_hdd", "key": "label"}
{"label": "Temperature #1 (WD_BLACK SN770 500GB)", "unit": "C", "value": 59.9,
 "warning": 83, "critical": 87, "type": "temperature_hdd", "key": "label"}
```

One expected side note: at idle two GPU sensors sit at CAREFUL (blue) — `GPU Core` 50 °C / `GPU Hot Spot` 61 °C — because GPU temperatures are typed `temperature_core` and inherit `temperature_core_careful=45` from the stock conf, and an idle RTX 3060 Ti simply lives around 50 °C. Not a bug in this PR (same story as any hot-running sensor vs the generic CPU thresholds), just mentioning it so nobody is surprised when reproducing.

The unit tests now cover each of these cases (15 tests for the grabber + 3 for the threshold alert logic). Screenshots of the web UI and TUI attached below.

final_tui_sensors
final_webui_sensors

Issue #3593, opened 2026-06-16
Add a new msgpack endpoint in the API


nicolargo:
For the moment, Glances expose an /all endpoint in the Restful API. This endpoint produce **an huge JSON file**.

The goal of the proposal is to add a new endpoint: /all/msgpack

That return the **Python Dict** with all plugins but packed by
 https://github.com/msgpack/msgpack-python.

Or better msgspec

Size of Python list with all plugins (before pack): 581.097 bytes
Size after pack: 424.674 bytes
Compression: **27%**

PR #3616, opened 2026-07-18
Add a new /all/msgpack endpoint in the API #3593

yogendrarau:
#### Description
Add a compact binary alternative to the /all endpoint, which returns a very large JSON payload. The new GET /api/4/all/msgpack endpoint returns the same dictionary of all plugin stats serialized to MessagePack using msgspec (~25% smaller than JSON on a typical payload).

- GlancesMsgpackResponse mirrors GlancesJSONResponse; msgspec encodes datetime as ISO 8601 strings, so the decoded payload matches /all.
- msgspec is optional: the endpoint returns HTTP 501 if it is missing, and it is added to the "web" optional-dependencies group.
- Add a RESTful test, document the endpoint, and regenerate openapi.json.

#### Resume

* Bug fix: no
* New feature: yes
* Fixed tickets: #3593


nicolargo:
Hi @yogendrarau 

first of all thanks for the PR.

As said in the https://github.com/nicolargo/glances/issues/3593 issue, the feature is targeted in the Glances v5 version. This version will break the API and integrate new features (like the /all/msgpack option). 

Sorry but i must reject the PR for the moment. Feel free to contribute again the Glances v5 will be opened to PR.

PR #3624, opened 2026-07-24
Fix get_ip_address() returning the last interface instead of the first

### LINKED ISSUE #3617: IP plugin displays wrong interface: outer loop in get_ip_address() never breaks
The get_ip_address() function in glances/globals.py is supposed to return the IP address of the first active (up, non-loopback) network interface, but due to a missing break in the outer loop, it actually returns the address of the last qualifying interface instead.

Code in question (glances/globals.py, get_ip_address()):

python
def get_ip_address(ipv6=False):
    """Get current IP address and netmask as a tuple."""
    family = socket.AF_INET6 if ipv6 else socket.AF_INET
    stats = psutil.net_if_stats()
    addrs = psutil.net_if_addrs()
    ip_address = None
    ip_netmask = None
    for interface, stat in stats.items():
        if stat.isup and interface != 'lo':
            if interface in addrs:
                for addr in addrs[interface]:
                    if addr.family == family:
                        ip_address = addr.address
                        ip_netmask = addr.netmask
                        break  # only breaks the inner loop
    return ip_address, ip_netmask

The inner break only exits the loop over addresses for a single interface — the outer loop over stats.items() keeps going and overwrites ip_address/ip_netmask for every subsequent up, non-loopback interface that has a matching address family. So the function silently returns whichever qualifying interface happens to be last in psutil.net_if_stats()'s dict order, not the first.

Impact

On any host with Docker (or other virtual bridge/veth interfaces) installed, this reliably causes the [ip] plugin to display the Docker bridge's internal IP (e.g. 172.18.0.1) instead of the real host IP (e.g. 192.168.0.150) on the dashboard, because the bridge/veth interfaces are isup=True and appear later in net_if_stats() than the physical NIC.

Steps to reproduce

Run Glances on a host with Docker installed and at least one running container (so a br-xxxx bridge interface exists and is up).
Open the Web UI or TUI dashboard.
Observe the IP field shows the Docker bridge address rather than the host's real LAN address.

Environment

Glances version: 4.5.5
PsUtil version: 7.2.2
OS: Ubuntu (systemd service, standard install, not containerized)

Example psutil.net_if_stats() order on affected host:

['lo', 'enp1s0', 'docker0', 'br-cd45b0d15fb9', 'vethd553473', ...]

enp1s0 is second (correct answer), but br-cd45b0d15fb9 (up, with an IPv4 address) comes later and overwrites it.

Suggested fix

Add a second break (or an early return) once a valid address is found, so the outer loop stops at the first qualifying interface:

python
        if stat.isup and interface != 'lo':
            if interface in addrs:
                for addr in addrs[interface]:
                    if addr.family == family:
                        ip_address = addr.address
                        ip_netmask = addr.netmask
                        break
                if ip_address:
                    break

I've tested this patch locally and it correctly resolves the private IP to the physical host interface instead of the Docker bridge.

Matcinis:
## What

`get_ip_address()` in `glances/globals.py` was returning the **last** matching interface's address instead of the **first** one, because the outer loop over `psutil.net_if_stats().items()` had no `break` — only the inner loop over addresses for a single interface did.

## Why it matters

On hosts with Docker (or any virtual bridge interface), this silently returns the bridge IP (e.g. `172.18.0.1`) instead of the real LAN address (e.g. `192.168.0.150`), whenever the virtual interface happens to sort after the real one in dict order.

## Fix

Break out of the outer loop as soon as a qualifying address has been found for an interface, so the function returns the first up/non-loopback interface instead of the last.

## Tests

Added `tests/test_globals.py` with a regression test that reproduces the exact Docker-bridge scenario from the issue (fails on the old code, passes after the fix — verified locally by temporarily reverting the fix), plus coverage for loopback/down-interface skipping and the no-match case.

Ran `ruff check` / `ruff format --check` on both changed files — no issues.

Fixes #3617

### DIFF
--- a/glances/globals.py
+++ b/glances/globals.py
@@ -761,6 +761,8 @@ def get_ip_address(ipv6=False):
                         ip_address = addr.address
                         ip_netmask = addr.netmask
                         break
+            if ip_address:
+                break
 
     return ip_address, ip_netmask
 
--- a/tests/test_globals.py
+++ b/tests/test_globals.py
@@ -1,0 +1,88 @@
+#!/usr/bin/env python
+#
+# Glances - An eye on your system
+#
+# SPDX-FileCopyrightText: 2026 Nicolas Hennion <nicolas@nicolargo.com>
+#
+# SPDX-License-Identifier: LGPL-3.0-only
+#
+
+"""Tests for glances.globals helper functions."""
+
+import socket
+from collections import OrderedDict
+from types import SimpleNamespace
+from unittest.mock import patch
+
+from glances.globals import get_ip_address
+
+
+def _stat(isup=True):
+    return SimpleNamespace(isup=isup, duplex=0, speed=0, mtu=1500, flags='')
+
+
+def _addr(family, address, netmask='255.255.255.0'):
+    return SimpleNamespace(family=family, address=address, netmask=netmask, broadcast=None, ptp=None)
+
+
+class TestGetIpAddress:
+    """get_ip_address() should return the FIRST up, non-loopback interface's address."""
+
+    def test_returns_first_matching_interface_not_last(self):
+        # Regression test for #3617: on hosts with Docker, a virtual bridge
+        # interface (docker0) sorted after the real LAN interface used to
+        # silently win because the outer loop never stopped scanning.
+        stats = OrderedDict(
+            [
+                ('eth0', _stat()),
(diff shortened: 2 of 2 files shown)


nicolargo:
Merged into develop.

Thanks @Matcinis  !

PR #3626, opened 2026-07-25
fix(exports): correct column/value mismatch for list plugins in TimescaleDB

### LINKED ISSUE #3592: Glances Network plugin with mismatched schema not logging to MacOS
**Check the bug**

Found similar bugs but old and not completely relevant to my issue
https://github.com/nicolargo/glances/issues/3014
https://github.com/nicolargo/glances/issues/3036

**Describe the bug**

network plugin isn't logging to my TimescaleDB PostgreSQL database on my macOS Mac Mini M4 running glances

**To Reproduce**

1. Start Glances
2. network plugin doesn't log data due to mismatched schema

**Expected behavior**

Data is logged to my PostgreSQL database

```
glances -V
Glances version:	4.5.4
Glances API version:	4
PsUtil version:		7.2.2
```

Installed through brew, managed by pm2

```
glances --issue
===============================================================================
Glances 4.5.4 (/opt/anaconda3/lib/python3.12/site-packages/glances/__init__.py)
Python 3.12.7 (/opt/anaconda3/bin/python)
PsUtil 7.2.2 (/opt/anaconda3/lib/python3.12/site-packages/psutil/__init__.py)
===============================================================================
...
network       [OK]    0.00293s
...
{'hostname': 'Mac-mini.local',
 'hr_name': 'Darwin 15.7.3 64bit',
 'os_name': 'Darwin',
 'os_version': '15.7.3',
 'platform': '64bit'}
```

**Additional context**

log errors:
> 2026-06-15 19:04:15,134 -- ERROR -- Cannot insert data into table network: the query has 25 placeholders but 24 parameters were passed

Haven't found anything regarding this

Sanjays2402:
#### Description

Closes #3592.

The list-plugin branch of the TimescaleDB exporter counts the `key` field once as a column but twice in the values: it is appended manually as `key_id`, then again via `.extend(plugin_item.values())`. The `item_list[:-1]` slice tries to compensate but drops the last stat field instead of `key`, and still leaves the count off by one — so an N-field plugin builds `3 + N` columns for `2 + N` values and the insert fails with "the query has 25 placeholders but 24 parameters were passed" (the reporter hit this on `network`).

This excludes `key` from both the column generation and the values extend (it is already stored as `key_id`) and drops the slice, so the counts match and no stat field is lost.

`tests/test_export_timescaledb_list.py` asserts the column/value counts match, that `key` appears only as `key_id`, and that the previously dropped trailing field is exported. All three fail on `develop` and pass with the fix.

#### Resume

* Bug fix: yes
* New feature: no
* Fixed tickets: 3592

### DIFF
--- a/glances/exports/glances_timescaledb/__init__.py
+++ b/glances/exports/glances_timescaledb/__init__.py
@@ -144,6 +144,9 @@ class Export(GlancesExport):
                 creation_list.append('key_id TEXT NOT NULL')
                 segmented_by.extend(['hostname_id', 'key_id'])  # Segment by hostname and key
                 for key, value in plugin_stats[0].items():
+                    # The 'key' field is already exported as key_id, do not duplicate it
+                    if key == 'key':
+                        continue
                     creation_list.append(f"{key} {convert_types[type(value).__name__]} NULL")
                 # Create the values list (it is a list of list to have a single datamodel for all the plugins)
                 for plugin_item in plugin_stats:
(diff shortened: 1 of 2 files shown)


nicolargo:
Merged into develop.

Thanks for the PR @Sanjays2402 !

Issue #3627, opened 2026-07-27
glances shows no warning when that temperature is high


slrslr:
Hello,

I am running Glances on [this hardware/system](https://justpaste.it/drjki).

the CPU temperature jumps from under 90C to over 95C for some reason despite CPU load does not change, but Glances does not indicate high temperature in orange or red text near "Tctl". For example I would expect red value when it is 95C

$ grep '\[Sensors\]' -A 16 ./.local/share/doc/glances/glances.conf

> [sensors]
> \# Documentation: https://glances.readthedocs.io/en/latest/aoa/sensors.html
> disable=False
> \# By default refresh every refresh time * 2
> \#refresh=6
> \# Hide some sensors
> \#hide=ambient
> \# Sensors core thresholds (in Celsius...)
> \# Default values are grabbed from the system
> \#temperature_core_careful=60
> \#temperature_core_warning=70
> **temperature_core_warning=90**
> \#temperature_core_critical=80
> **temperature_core_critical=95**
> \# Temperatures threshold in °C for hddtemp
> \# Default values if not defined: 45/52/60
> temperature_hdd_careful=45

My CPU official product page claims max. temperature 95C which was not present in glances.conf by default.

```
Package: glances
Version: 3.4.0.3+dfsg-1
Priority: extra
Section: universe/utils
Origin: Ubuntu
```
(Ubuntu 24 LTS based distribution repo)

[glances --issue](https://justpaste.it/e97dw)

Glances configuration file is [here](https://justpaste.it/du14v).
Glances log file is [here](https://justpaste.it/dl713).

nicolargo:
Hi @slrslr 

Glances 3.4.0.3 is an old version (more than 3 years).

 Please update it to latest (Glances 4.5.5) and test again.

Thanks !

slrslr:
I have updated this issue title and description. Please kindly check again if you can reproduce it on your current version.

yogendrarau:
Your config works on 4.5.5. I get `OK` at 85C, `WARNING` at 92C, `CRITICAL` at 95C.
Thresholds on a sensor **type** (`temperature_core_*`) only landed in 4.3.0 (#3049),
so on 3.4.0.3 that key is never read and Glances falls back to the system thresholds,
which `k10temp` doesn't provide for `Tctl`, hence no colour at all.

Workaround without upgrading: the per-sensor form does exist in 3.4.0.3 (#2058):

```ini
temperature_core_Tctl_careful=80
[...]

Issue #3049, opened 2024-12-21
Raise an (Alert) Event for a group of sensors


nicolargo:
### Discussed in https://github.com/nicolargo/glances/discussions/2692

<div type='discussions-op-text'>

<sup>Originally posted by **kt-alt** March  7, 2024</sup>
A few days ago, my server unexpectedly shutdown due to high temp.  I've installed glances monitor it and can see the temp readings, but it doesn't appear that an alert or event is raised for temperature.  

I have previously modified event.py to create syslog messages for events, which are sent over to graylog for centralized alerting.  I'd like to do this for my server temp as well.

Is there a way to enable this? Or is this something to suggest as a new feature.</div>

nicolargo:
Merged into develop.

nicolargo:
Test OK : https://github.com/nicolargo/glances/discussions/2692#discussioncomment-11642802

Issue #2058, opened 2022-05-29
In the sensor plugin configuration file limits should overwrite system ones


gauvins2:
**Is your feature request related to a problem? Please describe.**
Memory module sensor (jc42) values are displayed properly, but in red (critical)

**Describe the solution you'd like**
Being able to set values as we can for core (ex: temperature_core_careful=60)

**Additional context**
I've tried temperature_jc42_careful=50 and jc42_careful=50, to no avail

nicolargo:
Hi @gauvins2 

the temperature_jc42_careful=50 should work.

Are you sure that you read the good Glances configuration file ?

Please start Glances in debug mode (-d) and [paste/bin](https://pastebin.com/) the Glances log file.

gauvins2:
tried several variants of the sensors' names (jc42 / jc42* / jc42 1 / jc42[1-8]) but didn't work. Log shows that glances uses values of 0 for careful/warning/critical.

Log can be seen [here](https://pastebin.com/5Rpnsuxe)

nicolargo:
Your sensors are named: 
- jc42 1
- jc42 2
- ...
- jc42 8

So your configuration file should be:

```
[sensors]
temperature_jc42 1_careful=50
temperature_jc42 2_careful=50
temperature_jc42 3_careful=50
temperature_jc42 4_careful=50
temperature_jc42 5_careful=50
temperature_jc42 6_careful=50
temperature_jc42 7_careful=50
temperature_jc42 8_careful=50
```

gauvins2:
Hi again (and thanks for your time)

not working. See [here](https://pastebin.com/P28ZSBj8)

nicolargo:
Strange... perhaps the values configured in the Glances conf file is overwrite by the system one...

Can you please copy/paste the result of the following command:

```
python
>>> import psutil
>>> psutil.sensors_temperatures()
```

gauvins2:
as per your request: here <https://pastebin.com/58ehxy8c> (critical: 0 ...)

nicolargo:
Bug confirmed: if the system give a default value (0) then the values defined in the Glances conf file are not taken into account:

```
2022-07-14 09:44:31,963 -- DEBUG -- Load limit: sensors_temperature_acpitz 1_carful = 40.0
2022-07-14 09:44:31,963 -- DEBUG -- Load limit: sensors_temperature_acpitz 1_warning = 50.0
[...]

PR #3628, opened 2026-07-27
Sensors thresholds set in the configuration file are ignored when no critical level is defined #3627

yogendrarau:
#### Description

While investigating #3627 I could not reproduce the reported case on `develop`: the
reporter's configuration works correctly today.

```ini
[sensors]
temperature_core_warning=90
temperature_core_critical=95
```

| Tctl  | decoration |
|-------|------------|
| 85C   | OK         |
| 92C   | WARNING    |
| 95C   | CRITICAL   |

The reporter is on 3.4.0.3, which predates the sensor *type* thresholds added in
4.3.0 (#3049), so their `temperature_core_warning` key was never read on that
version. That part of the issue is a version problem, not a bug.

However, the dispatch that decides *which* thresholds to use has an adjacent bug
that produces exactly the same symptom on current `develop`. It only probes the
`critical` level:

```python
if self.is_limit('critical', stat_name=i['type'] + '_' + i['label']):
```

So a configuration that defines `careful` and/or `warning` but **no** `critical` is
silently discarded, and the plugin falls back to the system thresholds. This affects
both the per sensor (#2058) and the per type (#3049) forms.

The user visible consequence is the one described in #3627. AMD `k10temp` reports no
`high`/`critical` for `Tctl` on many boards, so `__get_system_thresholds()` returns
`DEFAULT` and the temperature is never highlighted:

```
config: temperature_core_warning=90   system: high=None, critical=None
  80C -> DEFAULT
  92C -> DEFAULT
  95C -> DEFAULT
  99C -> DEFAULT
```

That the configuration is genuinely ignored (rather than coincidentally agreeing with
the system) is visible when the two disagree, with `warning=90` in the config and
`high=70` from the system:

```
  80C -> WARNING     config(90) says OK, system(70) says WARNING
```

This PR checks every criticality level instead of only `critical`, so the documented
precedence (specific sensor, then sensor type, then system) applies as soon as the
user defines any threshold. With the fix, the two cases above return `WARNING` at
92C/95C/99C and `OK` at 80C.

Notes:

* No behaviour change for the shipped default configuration, which defines all three
  levels, nor for the fan speed, hdd temperature and battery sensors.
* No new configuration key, and `docs/aoa/sensors.rst` already documents the
  precedence this restores, so no documentation change.
* `NEWS.rst` left untouched, following the usual release process.
* 8 regression tests added in `tests/test_plugin_sensors.py`; 5 of them fail without
  the fix and the other 3 pin the unchanged system threshold fallback. Full sensors
  suite: 40 passed, 3 skipped (Linux only). `make lint` and `make format` are clean.

#### Resume

* Bug fix: yes
* New feature: no
* Fixed tickets: #3627

### DIFF
--- a/glances/plugins/sensors/__init__.py
+++ b/glances/plugins/sensors/__init__.py
@@ -189,6 +189,14 @@ class SensorsPlugin(GlancesPluginModel):
 
         return stats
 
+    def __has_user_thresholds(self, stat_name):
+        """Return True if the user defined at least one threshold for the given stat_name.
+
+        All the criticality levels are checked because the user is free to define only
+        some of them (for example a warning without a critical). See #3627.
+        """
+        return any(self.is_limit(level, stat_name=stat_name) for level in ('careful', 'warning', 'critical'))
+
     def __get_system_thresholds(self, sensor):
         """Return the alert level thanks to the system thresholds.
         Note: Only Warning (aka High) and Critical thresholds are used. Careful is not available.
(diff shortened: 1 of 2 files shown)


nicolargo:
Thanks for the patch @yogendrarau

Issue #3631, opened 2026-07-28
GPU plugin duplicates card name and omits N/A for unavailable metrics in multi-GPU view


williamqwu:
**Check the bug**

We've checked and found no existing report of this bug. Closest is #3590 which covers the `EBUSY` read failure itself, not the rendering. #3630 is a relevant yet different issue raised by me.

**Describe the bug**

When a GPU sensor cannot be read, `_msg_curse_multi()` skips the metric entirely instead of rendering `N/A`, and the row is corrupted: the card name is printed twice and the utilisation column disappears. We note that 1) displaying `N/A` is the plugin's long-standing behaviour, not a new convention; 2) this affects the multi-GPU detailed view only.

```
AMD GPUAMD GPU mem   0%
```

Two problems compound in `glances/plugins/gpu/__init__.py`, `_msg_curse_multi()`:

1. Each metric has its own `if gpu_stats.get('proc') is not None:` guard. When the value is `None` the whole segment is suppressed. `_format_value()` already maps `None` to `'N/A'`, and the code prior to d8b1648b always called it. Dropping the column hides exactly the condition an operator needs to see.
2. When the processor guard is skipped, `msg` still holds the card name, and the memory branch does `msg += f' mem {mem_msg}'`, so the name is re-emitted.

Regression from d8b1648b ("Improve Intel CPU name", 2026-02-19). Verified present verbatim in released v4.5.1 and v4.5.5, and on `develop` (d6943c26).

**To Reproduce**

1. Use an AMD card whose `gpu_busy_percent` cannot be read. On our node all 8 GPUs are fenced by a fatal RAS event, so the driver rejects every dynamic sensor read with `EBUSY`:

  ```
  $ cat /sys/class/drm/card1/device/gpu_busy_percent
  cat: /sys/class/drm/card1/device/gpu_busy_percent: Device or resource busy
  $ dmesg | grep RAS
  [drm:amdgpu_driver_open_kms [amdgpu]] *ERROR* RAS Intr triggered, device disabled!!
  ```

  `read_file()` correctly returns `None`, so `get_proc()` yields `None`.
2. Start Glances: `glances`
3. Look at the GPU plugin panel.

Hardware-independent reproduction:

```python
from glances.plugins.gpu import GpuPlugin

p = GpuPlugin()
p.stats = [
    {'key': 'gpu_id', 'gpu_id': 'amd0', 'name': 'AMD GPU', 'proc': None, 'mem': 0,
[...]

Issue #3630, opened 2026-07-28
GPU plugin duplicates the utilisation value and paints it with the memory colour in multi-GPU view


williamqwu:
**Check the bug**

We've checked and found no existing report of this bug.

**Describe the bug**

In the multi-GPU curses view, every GPU row prints its processor utilisation **twice**, and the second copy is decorated with the **memory** alert colour instead of the processor one.

The cause is in `glances/plugins/gpu/__init__.py`, `_msg_curse_multi()`. Each `curse_add_line()` emits an independently-coloured segment, but `msg` is accumulated with `+=` instead of being reassigned before the memory segment:

```python
msg = f' {proc_msg}'
ret.append(self.curse_add_line(msg, ...proc decoration...))
msg += f' mem {mem_msg}'          # <-- should be  msg = f' mem {mem_msg}'
ret.append(self.curse_add_line(msg, ...mem decoration...))
```

So the memory segment re-emits the processor value, coloured as memory.

This is a regression from d8b1648b ("Improve Intel CPU name", 2026-02-19), which split one correct single-segment line into three decorated segments in order to add per-metric colouring. Verified present verbatim in released v4.5.1 and v4.5.5, and on `develop` (d6943c26).

**To Reproduce**

1. Start Glances on a host with more than one GPU: `glances`
2. Look at the GPU plugin panel.

Hardware-independent reproduction:

```python
from glances.plugins.gpu import GpuPlugin

p = GpuPlugin()
p.stats = [
    {'key': 'gpu_id', 'gpu_id': 'amd0', 'name': 'AMD GPU', 'proc': 0, 'mem': 97,
     'temperature': None, 'fan_speed': None},
    {'key': 'gpu_id', 'gpu_id': 'amd1', 'name': 'AMD GPU', 'proc': 0, 'mem': 97,
     'temperature': None, 'fan_speed': None},
]
p.update_views()
ret = []
p._msg_curse_multi(ret)
print(''.join(s['msg'] for s in ret))
```

Actual (the leading blank line is the row separator emitted by `curse_new_line()`):

```

AMD GPU   0%   0% mem  97%
AMD GPU   0%   0% mem  97%
```

**Expected behavior**

```
AMD GPU   0% mem  97%
```

One utilisation value, decorated with the processor alert level; one memory value, decorated with the memory alert level.

This matches the per-GPU view shown in the documentation ([`_images/pergpu.png`](https://glances.readthedocs.io/en/latest/_images/pergpu.png), linked from
[...]

Issue #3590, opened 2026-06-15
AMD GPU plugin crashes glances when the card is runtime-suspended (EBUSY not caught)


kwehden:
## Description

The AMD GPU plugin (`glances/plugins/gpu/cards/amd.py`) crashes glances when the GPU is runtime power-managed and currently suspended. On a discrete AMD card with `power/control = auto`, the amdgpu driver does **not** wake the card for sysfs sensor reads — instead it returns `OSError(errno 16, EBUSY, "Device or resource busy")`.

The plugin guards those reads with `except PermissionError:` only (added in #3125). `PermissionError` maps to `EACCES`/`EPERM`, so `EBUSY` is **not** caught, propagates out of the plugin, and takes down the whole UI.

## Reproduce

On a machine with a discrete AMD GPU using runtime PM, while the card is idle/suspended:

```
$ cat /sys/class/drm/card0/device/power/runtime_status
suspended
$ cat /sys/class/drm/card0/device/power/control
auto
$ cat /sys/class/drm/card0/device/gpu_busy_percent
cat: ...gpu_busy_percent: Device or resource busy
```

Then run glances — it errors out as soon as the GPU plugin samples the suspended card.

Minimal confirmation that it's not `PermissionError`:

```python
>>> open("/sys/class/drm/card0/device/gpu_busy_percent").read()
OSError: [Errno 16] Device or resource busy   # isinstance(e, PermissionError) == False
```

(Note: `mem_info_vram_used`/`mem_info_vram_total` happen to still read fine while suspended; it's `gpu_busy_percent` in `get_proc()` and `temp*_input` in `get_temperature()` that raise.)

## Environment

- Glances 4.3.3
- Python 3.14.4, psutil 7.1.0
- OS: Ubuntu 26.04 LTS
- Kernel: Linux 7.0.0-22-generic x86_64
- GPU: AMD Radeon RX 6600 — Navi 23 [`1002:73ff`] (rev c7)
- Driver: amdgpu (kernel driver in use: amdgpu), runtime PM enabled (`power/control = auto`)

## Suggested fix

Broaden the exception handling in `get_mem()`, `get_proc()`, and `get_temperature()` from `except PermissionError:` to `except OSError:` (`PermissionError` is a subclass of `OSError`, so this still covers #3125 while also handling `EBUSY` from a suspended card). The affected functions can simply return `None` as they already do — sensors then show as unavailable while the card sleeps and repopulate when it wakes.

```diff
-            except PermissionError:
[...]

PR #3633, opened 2026-07-28
Fix duplicated and suppressed metrics in the multi-GPU curses view

williamqwu:
#### Description

Relevant issue: [#3630](https://github.com/nicolargo/glances/issues/3630) and [#3631](https://github.com/nicolargo/glances/issues/3631).

The multi-GPU curses view renders each GPU row as several `curse_add_line()` segments so that the processor and memory values can carry their own alert decoration. Two defects were introduced when that split was made in d8b1648b, and they interact.

First, the local `msg` is accumulated with `+=` rather than reassigned between the two segments. `curse_add_line()` emits each segment independently, so the memory segment re-emits whatever the processor segment already printed, and repaints it with the memory alert colour. A row showing an idle GPU with high VRAM pressure renders as `AMD GPU   0%   0% mem  97%`. The utilisation value appears twice, and the second copy is red because memory is critical.

Second, each segment is guarded by `if gpu_stats.get(...) is not None:`, which suppresses the column entirely when a sensor cannot be read. `_format_value()` already maps `None` to `'N/A'`, and the summary view (`_msg_curse_summary()`) relies on exactly that. Suppressing the column instead is a loss of information in the one situation where it matters most, and it also leaves `msg` holding the card name, so the `+=` above duplicates the name: `AMD GPUAMD GPU mem   0%`.

Unreadable sensors are not an edge case: amdgpu returns `EBUSY` for `gpu_busy_percent` while a card is runtime-suspended (#3590) or fenced by a fatal RAS event, NVIDIA returns `None` on `NVMLError`, and ARM's `proc` is `None` on its first sample by design. `_msg_curse_multi()` is shared by every backend, so neither defect is vendor-specific; this surfaced on an 8x AMD Instinct MI355X host whose GPUs were all RAS-disabled, but a multi-NVIDIA host shows the duplication identically.

The fix drops the `is not None` guards and always emits both segments through `_format_value()`, building each segment string independently. Behaviour on healthy hardware is unchanged apart from the removal of the duplicated value; unavailable sensors now read `N/A`, consistent with the summary view.

Regression range: introduced in d8b1648b; present verbatim in released v4.5.1 and v4.5.5.

Before:

```
AMD GPU   0%   0% mem  97%      # healthy card, utilisation printed twice
AMD GPUAMD GPU mem   0%         # sensor unreadable, name printed twice, no utilisation column
```

After:

```
AMD GPU   0% mem  97%
AMD GPU  N/A mem   0%
```

The "after" form matches the per-GPU view in the documentation ([`_images/pergpu.png`](https://glances.readthedocs.io/en/latest/_images/pergpu.png)), which shows
one utilisation value per row.

#### Resume

* Bug fix: yes
* New feature: no

### DIFF
--- a/glances/plugins/gpu/__init__.py
+++ b/glances/plugins/gpu/__init__.py
@@ -332,7 +332,7 @@ class GpuPlugin(GlancesPluginModel):
                 )
             if gpu_stats.get('mem') is not None:
                 mem_msg = self._format_value(gpu_stats.get('mem'))
-                msg += f' mem {mem_msg}'
+                msg = f' mem {mem_msg}'
                 ret.append(
                     self.curse_add_line(
                         msg,
--- a/glances/plugins/gpu/__init__.py
+++ b/glances/plugins/gpu/__init__.py
@@ -319,26 +319,21 @@ class GpuPlugin(GlancesPluginModel):
             ret.append(self.curse_new_line())
             # id_msg = '{:7}'.format(gpu_stats['gpu_id'])
             id_msg = '{:7}'.format(gpu_stats['name'][0:9])
-            msg = f'{id_msg}'
-            ret.append(self.curse_add_line(msg))
-            if gpu_stats.get('proc') is not None:
-                proc_msg = self._format_value(gpu_stats.get('proc'))
-                msg = f' {proc_msg}'
-                ret.append(
-                    self.curse_add_line(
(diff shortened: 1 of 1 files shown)


nicolargo:
Thanks for the PR (and issue reports) @williamqwu  !

Just merged into develop.

PR #3635, opened 2026-07-30
fix(ports): resolve alert level by severity instead of dict ordering

### LINKED ISSUE #3632: Alert level decided by dict ordering: a failing+slow URL is downgraded to WARNING, an unscanned URL reports CRITICAL
**Describe the bug**

`PortsPlugin.get_default_ret_value()` collapses all matching alert conditions into a single dict key, so the level that "wins" is whichever condition happens to be evaluated **last**, not the most severe one:

```python
def get_default_ret_value(self, conds):
    ret_as_dict_val = {'ret': key for key, cond in conds.items() if cond}
    return ret_as_dict_val.get('ret', 'OK')
```

Every truthy condition overwrites the same `'ret'` key, so the result depends entirely on the insertion order of the dict built by `get_conds_if_url()` / `get_conds_if_port()` (`CAREFUL`, `CRITICAL`, `WARNING`).

Two consequences for the web/URL checks:

1. **A failing URL that is also slow is downgraded to WARNING.** `WARNING` is evaluated after `CRITICAL`, so when a URL returns a non-2xx/3xx code *and* takes longer than `web_N_rtt_warning`, the `CRITICAL` verdict is overwritten. A site that is both broken and slow — the common signature of an overloaded or failing backend — is reported as less serious than the same site returning 404 quickly. This is permanent, not a startup artifact: it persists for as long as both conditions hold.

2. **A URL that has not been scanned yet is reported CRITICAL.** `status is None` satisfies `CAREFUL` (intended: "Scanning") but *also* satisfies `CRITICAL`, because `None not in [200, 301, 302]` is `True`. So between Glances starting and the scanner thread completing its first pass, every configured URL shows as CRITICAL rather than "Scanning".

Point 2 also reaches the action system. `get_p_alert()` calls `manage_action(stat_name, ret.lower(), ...)`, and `get_limit_action()` falls back to the plugin-wide `ports_critical_action` key, so a `ports_critical_action` configured in `glances.conf` can be executed for a URL that is merely still being scanned. This one is timing-dependent: `GlancesActions.run()` suppresses actions for the first `args.time * 2` seconds (4 s by default, issue #732), so it fires only when the first web scan finishes after that window. That is easy to hit — `ThreadScanner.run()` scans the `[ports]` host/port entries first with a `time.sleep(1)` between each, so a handful of configured ports pushes the first URL scan well past 4 s.

The host/port checks in `get_conds_if_port()` are not currently affected, but only by luck: their conditions happen to be mutually exclusive for reachable values. The ordering dependence is the same.

**To Reproduce**

The alert level is a pure function of the scan-result dict, so no network access or config file is needed:

```python
from glances.plugins.ports import PortsPlugin

plugin = PortsPlugin.__new__(PortsPlugin)

cases = [
    ("not scanned yet",      {'status': None, 'elapsed': 0,   'rtt_warning': None}, 'CAREFUL'),
    ("404 AND slow",         {'status': 404,  'elapsed': 2.0, 'rtt_warning': 1.0},  'CRITICAL'),
    ("404, fast (control)",  {'status': 404,  'elapsed': 0.1, 'rtt_warning': 1.0},  'CRITICAL'),
    ("200 but slow (ctrl)",  {'status': 200,  'elapsed': 2.0, 'rtt_warning': 1.0},  'WARNING'),
]

for label, web, expected in cases:
    actual = plugin.get_default_ret_value(plugin.get_conds_if_url(web))
    print(f"[{'ok ' if actual == expected else 'BUG'}] {label}: got {actual}, expected {expected}")
```

Output:

```
[BUG] not scanned yet: got CRITICAL, expected CAREFUL
[BUG] 404 AND slow: got WARNING, expected CRITICAL
[ok ] 404, fast (control): got CRITICAL, expected CRITICAL
[ok ] 200 but slow (ctrl): got WARNING, expected WARNING
```

To see it end to end instead, add to `glances.conf` and start Glances — the entry shows red before the first scan completes, then settles on an orange `Code 404` once the scan returns:

```ini
[ports]
refresh=30
web_1_url=https://httpstat.us/404?sleep=2000
web_1_rtt_warning=1000
```

**Expected behavior**

- A URL returning a non-2xx/3xx code should report `CRITICAL` regardless of its response time.
- A URL whose first scan has not completed should report `CAREFUL` ("Scanning"), and should not trigger `ports_critical_action`.
- More generally, when several conditions match, the most severe should win, independent of dict ordering.

**Suggested fix**

Two changes are needed; either one alone leaves the bug half-present.

Resolve by severity rather than by iteration order:

```python
def get_default_ret_value(self, conds):
    for level in ('CRITICAL', 'WARNING', 'CAREFUL'):
        if conds.get(level):
            return level
    return 'OK'
```

and stop `None` from matching the `CRITICAL` condition, since ordering alone would still rank a not-yet-scanned URL as CRITICAL (both `CAREFUL` and `CRITICAL` are true for it):

```python
'CRITICAL': web['status'] is not None and web['status'] not in [200, 301, 302],
```

Happy to open a PR with both changes plus a unit test over the four cases above — the current behaviour fails silently, so there is nothing guarding it today.

**Environment**

- Operating System: Windows 11 (10.0.26200) — the defect is platform-independent, it is pure dict-ordering logic
- Glances & psutil versions: `Glances v4.5.6_dev01 with psutil v7.2.2`
- How do you install Glances: from source (`develop` branch)

Sanjays2402:
Closes #3632

#### Description

`PortsPlugin.get_default_ret_value()` built `{'ret': key for key, cond in conds.items() if cond}`, so every matching condition overwrote the same key and the level that won was whichever was evaluated last, not the most severe. Two consequences for the web/URL checks: a URL that is both failing and slow was reported WARNING instead of CRITICAL, and a URL whose first scan had not completed matched both CAREFUL and CRITICAL (because `None not in [200, 301, 302]`) and showed as CRITICAL, which can also fire `ports_critical_action`.

The fix resolves by severity (CRITICAL > WARNING > CAREFUL > OK) and stops a `None` status from matching the CRITICAL condition, as suggested in the issue. `get_conds_if_port()` is unchanged and its behaviour is unaffected.

Added `tests/test_plugin_ports.py` covering the severity ordering and the four web scan outcomes; it fails on the current code (3 failures) and passes with the fix. This change was prepared with AI assistance; the regression test was run locally and fails without the fix.

#### Resume

* Bug fix: yes
* New feature: no
* Fixed tickets: 3632

### DIFF
--- a/glances/plugins/ports/__init__.py
+++ b/glances/plugins/ports/__init__.py
@@ -138,7 +138,7 @@ class PortsPlugin(GlancesPluginModel):
     def get_conds_if_url(self, web):
         return {
             'CAREFUL': web['status'] is None,
-            'CRITICAL': web['status'] not in [200, 301, 302],
+            'CRITICAL': web['status'] is not None and web['status'] not in [200, 301, 302],
             'WARNING': web['rtt_warning'] is not None and web['elapsed'] > web['rtt_warning'],
         }
 
@@ -162,9 +162,13 @@ class PortsPlugin(GlancesPluginModel):
         return ret
 
     def get_default_ret_value(self, conds):
-        ret_as_dict_val = {'ret': key for key, cond in conds.items() if cond}
+        # Resolve by severity, not by dict insertion order: several conditions
+        # can match at once and the most severe one has to win.
+        for level in ('CRITICAL', 'WARNING', 'CAREFUL'):
+            if conds.get(level):
+                return level
 
-        return ret_as_dict_val.get('ret', 'OK')
+        return 'OK'
 
     def set_status_if_host(self, p):
         if p['host'] is None:
--- a/tests/test_plugin_ports.py
+++ b/tests/test_plugin_ports.py
@@ -1,0 +1,67 @@
+#!/usr/bin/env python
+#
+# Glances - An eye on your system
+#
+# SPDX-FileCopyrightText: 2026 Nicolas Hennion <nicolas@nicolargo.com>
+#
+# SPDX-License-Identifier: LGPL-3.0-only
+#
+
+"""Tests for the Ports plugin."""
+
+import pytest
+
+from glances.plugins.ports import PortsPlugin
+
+
+@pytest.fixture
+def ports_plugin():
+    """Return a Ports plugin instance without running its full init."""
+    return PortsPlugin.__new__(PortsPlugin)
+
+
+def web_scan(status, elapsed=0, rtt_warning=1):
+    """Return a web scan result as stored by the ports plugin."""
+    return {'status': status, 'elapsed': elapsed, 'rtt_warning': rtt_warning}
+
+
+class TestPortsPluginAlertLevel:
+    """Test that the alert level is resolved by severity, not by dict ordering."""
+
+    @pytest.mark.parametrize(
+        ('conds', 'expected'),
+        [
+            ({'CAREFUL': True, 'CRITICAL': True, 'WARNING': True}, 'CRITICAL'),
+            ({'CAREFUL': True, 'CRITICAL': False, 'WARNING': True}, 'WARNING'),
+            ({'CAREFUL': True, 'CRITICAL': False, 'WARNING': False}, 'CAREFUL'),
+            ({'CAREFUL': False, 'CRITICAL': False, 'WARNING': False}, 'OK'),
+        ],
+    )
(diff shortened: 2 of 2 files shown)


nicolargo:
Many thanks for the PR @Sanjays2402 

Just merged into develop, will be release in targeted milestone version.

Issue #2749, opened 2024-05-02
Improve Github action unit test on BSD systems


nicolargo:
Like is done here: https://github.com/giampaolo/psutil/blob/master/.github/workflows/bsd.yml

Do not forget to fix the *BSD version, example for FreeBSD:
- https://github.com/vmactions/freebsd-vm?tab=readme-ov-file
- https://github.com/vmactions/freebsd-vm/tree/main/conf

nicolargo:
For FreeBSD:
- python3
- py-pip

PR #3636, opened 2026-08-02
Run the unit tests on FreeBSD

neilpang:
#### Description

The `test-freebsd` job has been commented out since it was written, and as
written it could not have worked:

- It used two separate `vmactions` steps, one to install dependencies and one
  to run pytest. Each step boots its own VM, so pytest ran in a VM where the
  dependencies had never been installed.
- It carried an `actions/setup-python` step and a `python-version` matrix,
  which configure the Linux host and have no effect inside the guest.

This folds it into a single step with `prepare`/`run`, drops the host-side
Python setup, and caches the prepared VM image.

It also installs only the runtime requirements. `dev-requirements.txt` pulls
in matplotlib's build chain (contourpy, cffi) and FreeBSD ships no wheels for
those, so pip builds them from source -- that alone ran past the job timeout.
pytest and psutil come from packages instead.

**Verification**

https://github.com/neilpang/glances/actions/runs/30731827218

```
Python 3.12.13 on FreeBSD 15.1
py312-pytest 8.4.2, py312-psutil 7.2.2
======================== 48 passed, 4 skipped in 7.93s =========================
```

Job time 2m50s. With `cache-after-prepare` the package install is skipped on
subsequent runs.

#### Resume

* Bug fix: no
* New feature: yes -- restores FreeBSD test coverage
* Fixed tickets: #2749

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -135,37 +135,38 @@ jobs:
         run: |
           python -m pytest ./tests/test_core.py
 
-  # test-freebsd:
+  test-freebsd:
 
-  #   needs: source-code-checks
-  #   runs-on: ubuntu-latest
-  #   strategy:
-  #     matrix:
-  #       # Only test the latest stable version
-  #       python-version: ["3.14"]
-  #   steps:
+    needs: source-code-checks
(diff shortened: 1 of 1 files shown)


nicolargo:
Thanks for the PR @neilpang 

Issue #3221, opened 2025-07-03
Client/server mode  not compatible with process cursor navigation


nicolargo:
Start a Glances server

Start a Glances client

Hit bottom key

Crash with following log:

2025-07-03 14:14:42,232 -- CRITICAL -- unsupported operand type(s) for -: 'NoneType' and 'int'

nicolargo:
See in Glances 4.3.2_rc01

nicolargo:
Crash corrected. 

Cursor disable in client/server mode.

Let the issue open to implement it in future release.

PR #3642, opened 2026-08-04
Fix cursor navigation in client/server mode (#3221)

art22s:
## Description

Client/server mode disables process cursor navigation, but the left/right process-name scroll handlers were not subject to the same guard as the up/down and kill handlers. Pressing those keys could still mutate cursor state even though cursor navigation is unavailable in client mode.

This change also makes `processes_count` safe when no process display limit has been assigned, preventing the `NoneType - int` crash reported by the issue.

## Verification

- `python3 -m pytest tests/test_glances_curses.py tests/test_plugin_processcount.py` — 46 passed
- `python3 -m pytest ./tests/test_core.py` — 52 passed
- `ruff check` — passed
- `ruff format --check` on changed files — passed

The available full suite excluding the optional Selenium-dependent WebUI module ran 713 passed, with four unrelated environment/baseline failures.

## Resume

* Bug fix: yes
* New feature: no
* Fixed tickets: #3221

---

### DIFF
--- a/glances/outputs/glances_curses.py
+++ b/glances/outputs/glances_curses.py
@@ -277,10 +277,10 @@ class _GlancesCurses:
             in {curses.KEY_LEFT if self.args.arrow_keys_sort else curses.KEY_SLEFT}: self._handle_sort_left,
             self.pressedkey
             in {curses.KEY_RIGHT if self.args.arrow_keys_sort else curses.KEY_SRIGHT}: self._handle_sort_right,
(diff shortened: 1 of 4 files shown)


nicolargo:
@art22s thanks for the PR. Just merged into develop.

For the record, it does not correct the https://github.com/nicolargo/glances/issues/3221. The feature describe in the #3221 issue is to navigate through processes, not between columns.

PR #3646, opened 2026-08-05
fix: count terminated connection states instead of counting initiated twice

Dipet:
#### Description

`update_for_net_connections_method()` loops over `self.initiated_states` twice. The second loop is meant to fill the terminated counters, so `self.terminated_states` is defined but never used: `stats['terminated']` always equals `stats['initiated']`, and `FIN_WAIT1`, `FIN_WAIT2`, `TIME_WAIT`, `CLOSE`, `CLOSE_WAIT` and `LAST_ACK` are never reported at all.

On a live host:

```
terminated_states: [FIN_WAIT1, FIN_WAIT2, TIME_WAIT, CLOSE, CLOSE_WAIT, LAST_ACK]   <- defined
keys produced    : [ESTABLISHED, LISTEN, SYN_RECV, SYN_SENT, 'initiated', 'terminated']
terminated states never set: all six
initiated=2  terminated=2  equal=True
```

The same machine had 556 connections in `TIME_WAIT` and 157 in `CLOSE_WAIT` at that moment, none of which reached the UI.

One word changed. Adds `tests/test_connections_states.py`, which feeds the plugin a fixed set of connections and checks the counters; two of its four tests fail on current `develop` and pass with the fix.

#### Resume

* Bug fix: yes
* New feature: no
* Fixed tickets:

### DIFF
--- a/glances/plugins/connections/__init__.py
+++ b/glances/plugins/connections/__init__.py
@@ -120,7 +120,7 @@ class ConnectionsPlugin(GlancesPluginModel):
             initiated += stats[s]
         stats['initiated'] = initiated
         terminated = 0
-        for s in self.initiated_states:
+        for s in self.terminated_states:
             stats[s] = len([c for c in net_connections if c.status == s])
             terminated += stats[s]
         stats['terminated'] = terminated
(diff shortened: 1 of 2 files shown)


nicolargo:
You are right !

Thanks for the PR @Dipet , just merged into develop.

PR #3647, opened 2026-08-05
fix: make the TTL cache actually cache

Dipet:
#### Description

Two defects in the `weak_lru_cache` / `_get_ttl_hash` pair. They have to be fixed together, because the second one currently hides the first.

**1. The TTL is ignored.** `_get_ttl_hash()` returns `datetime.now().second` and never looks at the `ttl` argument beyond the `None` check. So `ttl=30` behaves exactly like `ttl=1`, and because the value cycles every minute, an entry created at second 17 collides with second 17 of the *next* minute — a cache hit on data that is 60 seconds old. The docstring already says `TODO: Implement a real TTL mechanism`.

Now bucketed on a monotonic clock: the key stays constant for `ttl` seconds and never repeats.

**2. The cache holds one entry while the working set is dozens.** Both call sites left `maxsize` at its default of 1, and both are called once per plugin in a row. Every call evicts the previous entry, so nothing is ever reused:

```
processlist    0.0ms     (cached)
cpu            0.4ms     <- evicts processlist
processlist  207.2ms     <- recomputed, same second, ttl=1
```

With 35 plugins the cache never hits at all. After the fix, a full sweep of every plugin inside one TTL window costs 0.0 ms instead of 573 ms, and still recomputes once the window closes:

```
sweep 1 (same second):      0.0ms
sweep 2 (same second):      0.0ms
sweep 3 (same second):      0.0ms
sweep after 1.2s:        3173.4ms
```

#### On memory

`update_plugin()` returns `None`, so the cache stores keys only — a weak ref, a plugin name and an int, roughly 25 KB at `maxsize=128`. `GlancesApi.__getattr__` stores a reference to an already-live plugin object, not a copy.

`maxsize` cannot be dropped to `None`: with bucketed keys the entries of expired buckets stay until evicted, so an unbounded cache would grow forever. The bound is deliberate.

#### Behaviour change worth flagging

`GlancesApi.__getattr__` is decorated with `ttl=ttl`, where `ttl` is the refresh interval. Until now that value was ignored and the effective window was always under a second; it is now honoured. If a shorter window is wanted there, say so and I will pin it.

#### Tests

Adds `tests/test_ttl_cache.py` covering: the value is stable inside the window, changes after it, honours the requested `ttl`, and does not repeat after a minute. Four of its five tests cannot pass on current `develop`.

Full suite on Linux (web tests skipped, no FastAPI/selenium in this environment): **642 passed, 4 failed**. All four — `test_perf_update`, `test_memoryleak_no_history` and two in `test_actions_sanitize` — fail identically on unmodified `develop` on this machine. `test_perf_update` asserts more than one update per second, which this host (15 000 processes) cannot do either way.

#### Resume

* Bug fix: yes
* New feature: no
* Fixed tickets:

### DIFF
--- a/glances/api.py
+++ b/glances/api.py
@@ -35,7 +35,9 @@ class GlancesAPI:
         # Init the stats of all plugins in order to ensure that rate are computed
         self._stats.update()
 
-    @weak_lru_cache(maxsize=1, ttl=ttl)
+    # maxsize has to exceed the number of plugins: callers walk several of them in a row,
+    # so a smaller cache would evict every entry before it could ever be reused.
+    @weak_lru_cache(maxsize=128, ttl=ttl)
     def __getattr__(self, item):
         """Fallback to the stats object for any missing attributes."""
         if item in self._stats.getPluginsList():
--- a/glances/globals.py
+++ b/glances/globals.py
@@ -25,6 +25,7 @@ import re
 import socket
 import subprocess
 import sys
+import time
 import weakref
 from collections import OrderedDict
 from configparser import ConfigParser, NoOptionError, NoSectionError
@@ -554,13 +555,16 @@ def folder_size(path, errno=0):
 
 
 def _get_ttl_hash(ttl):
-    """A simple (dummy) function to return a hash based on the current second.
-    TODO: Implement a real TTL mechanism.
+    """Return a value that stays constant for ttl seconds, then changes.
+
+    Used as part of an lru_cache key, so that an entry stops being reused once it is older
+    than ttl. Buckets are aligned to a monotonic clock: an entry created just before a
+    boundary lives less than the full ttl, which is the price of expiring through the cache
+    key instead of storing a timestamp per entry.
     """
     if ttl is None:
         return 0
-    now = datetime.now()
-    return now.second
+    return int(time.monotonic() / ttl)
 
 
 def weak_lru_cache(maxsize=1, typed=False, ttl=None):
--- a/glances/stats.py
+++ b/glances/stats.py
@@ -311,9 +311,11 @@ please rename it to "{plugin_path.capitalize()}Plugin"'
         for p in self.getPluginsList(enable=False):
             self._plugins[p].load_limits(config)
 
-    # It's a weak cache to avoid updating the same plugin too often
+    # It's a weak cache to avoid updating the same plugin too often.
+    # maxsize has to exceed the number of plugins: this is called once per plugin in a row,
+    # so a smaller cache would evict every entry before it could ever be reused.
     # Note: the function always return None
-    @weak_lru_cache(ttl=1)
+    @weak_lru_cache(maxsize=128, ttl=1)
     def update_plugin(self, p):
         """Update stats, history and views for the given plugin name p"""
         self._plugins[p].update()
--- a/tests/test_ttl_cache.py
+++ b/tests/test_ttl_cache.py
@@ -1,0 +1,48 @@
+"""Tests for the TTL-based lru cache helpers in glances.globals."""
+
+from unittest import mock
+
+import pytest
+
+from glances.globals import _get_ttl_hash
+
+
+@pytest.fixture
+def clock():
+    with mock.patch('glances.globals.time.monotonic') as m:
+        yield m
+
+
+def test_hash_is_stable_within_the_ttl(clock):
+    clock.return_value = 1000.0
+    first = _get_ttl_hash(5)
+    clock.return_value = 1004.0
+    assert _get_ttl_hash(5) == first
+
+
+def test_hash_changes_once_the_ttl_has_passed(clock):
+    clock.return_value = 1000.0
+    first = _get_ttl_hash(5)
+    clock.return_value = 1006.0
+    assert _get_ttl_hash(5) != first
+
+
+def test_hash_honours_the_requested_ttl(clock):
+    """A longer ttl must hold its value where a shorter one has already moved on."""
+    clock.return_value = 1000.0
+    short, long = _get_ttl_hash(1), _get_ttl_hash(30)
+    clock.return_value = 1002.0
+    assert _get_ttl_hash(1) != short
+    assert _get_ttl_hash(30) == long
+
+
+def test_hash_does_not_repeat_after_a_minute(clock):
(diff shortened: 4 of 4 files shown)


nicolargo:
Hi @Dipet 

This one is tricky. There is a lot of issues / enhancements regarding CPU and memory consumption on this topic.

For the moment, your PR did not pass the perf test (make test-perf):

```
❯ make test-perf
.venv-uv/bin/uv run pytest tests/test_perf.py
=========================================================================== test session starts ============================================================================
platform linux -- Python 3.14.0, pytest-9.1.1, pluggy-1.6.0
rootdir: /home/nicolargo/dev/glances
configfile: pyproject.toml
plugins: asyncio-1.4.0, anyio-4.14.2
asyncio: mode=Mode.STRICT, debug=False, asyncio_default_fixture_loop_scope=None, asyncio_default_test_loop_scope=function
collected 1 item

tests/test_perf.py F                                                                                                                                                 [100%]

================================================================================= FAILURES =================================================================================
_____________________________________________________________________________ test_perf_update _____________________________________________________________________________

glances_stats = <glances.stats.GlancesStats object at 0x783f0a807380>

    def test_perf_update(glances_stats):
        """
        Test Glances perf.
        """
        test_duration = 12  # seconds
        perf_timer = Timer(test_duration)
        counter = 0
        from_cache = 0
        from_update = 0
        previous_interrupts_gauge = None
        while not perf_timer.finished():
            glances_stats.update()
            # interrupts_gauge should always increase
            interrupts_gauge = glances_stats.get_plugin('cpu').get_raw().get('interrupts_gauge')
            if interrupts_gauge is not None:
                if interrupts_gauge == previous_interrupts_gauge:
                    from_cache += 1
                else:
                    from_update += 1
                    previous_interrupts_gauge = interrupts_gauge
            counter += 1
        print(f"{counter} iterations. From cache: {from_cache} | From update: {from_update}")
        assert counter > test_duration
        assert from_update < from_cache
        assert from_cache >= test_duration * 2
>       assert from_update >= (test_duration / 2) - 1
E       assert 3 >= ((12 / 2) - 1)

tests/test_perf.py:40: AssertionError
---------------------------------------------------------------------------- Captured log setup ----------------------------------------------------------------------------
INFO     root:config.py:210 Read configuration file './conf/glances.conf'
ERROR    root:lxd.py:233 containers (LXD) plugin - Can't connect to LXD (('Connection aborted.', FileNotFoundError(2, 'No such file or directory')))
INFO     root:__init__.py:231 Followings processes stats will not be displayed: cpu_num
--------------------------------------------------------------------------- Captured stdout call ---------------------------------------------------------------------------
655720 iterations. From cache: 500734 | From update: 3
========================================================================= short test summary info ==========================================================================
FAILED tests/test_perf.py::test_perf_update - assert 3 >= ((12 / 2) - 1)
============================================================================ 1 failed in 12.56s ============================================================================
make: *** [Makefile:112: test-perf] Error 1
```

Dipet:
> There is a lot of issues / enhancements regarding CPU and memory consumption on this topic

Is increasing the LRU `max_size` a hard blocker due to strict memory constraints?
If the memory overhead is unacceptable, we might need to look into optimizing the cached entry size or adjusting the eviction logic instead.

nicolargo:
Yes  CPU and memory consumption are to be taken into account in priority for this kind of system monitoring tool.

The only goal of the TTL cache is to avoid a CPU consumption burst if a client asks stats at high bit-rate (several time per second). The current implementation limit update to one time per second witch is the minimal refresh rate configurable in Glances.

Dipet:
Thanks for the perf test — it caught a real regression, and the answer to your question is that memory is not the blocker here. The blocker is behavioural, and the test is right to fail.

## Why it fails

Making the `update_plugin` cache effective quantizes every plugin's refresh onto the TTL grid. `_check_decorator` resets `refresh_timer` only when an update actually runs, so the plugin's period is "at least `get_refresh()` **and** somebody called us". Today `update_plugin` is called at whatever rate the client polls, so the timer fires the instant it expires. With a working 1-second cache, the timer is only *examined* on bucket boundaries, and a 2-second refresh lands mid-bucket and waits for the next one.

Measured on the cpu plugin, `refresh = 2.0s`, intervals between actual refreshes:

```
maxsize=1 (develop)    [2.04, 2.04, 2.04, 2.04, 2.05, 2.04, 2.04, 2.04]
maxsize=128 (this PR)  [3.07, 3.03, 2.89, 2.01, 3.00, 3.00]
```

2 s becomes 3 s. That is exactly what `from_update` counts, hence `3 >= 5` failing on your box.

## What I tried instead

Since `update()` is already throttled by `_check_decorator` and `update_views()` / `update_stats_history()` are not, I tried recording in the decorator whether the stats were actually refreshed, and skipping the derived work when they were not. That passes the perf test and keeps the intervals at 2 s, but it does not pay off, for a reason worth reporting:

**`processlist` and `programlist` do not use `_check_decorator` at all.** Most plugins do; these two — the expensive ones — do not, so the flag never trips for them.

## Where the cost actually is

On a host with ~16 000 processes, per poll:

```
processlist   update()=0.1ms   update_views()=216ms
programlist   update()=153ms   update_views()=17ms
```

`processlist.update()` does not collect anything — it takes the list `processcount` already built. But `update_views()` rebuilds decorations for every field of every process on every poll, and `programlist.update()` re-runs `processes_to_programs()` over the same input.

The signal to skip both is already there and costs nothing: `glances_processes.get_list()` returns the *same list object* until `processcount` rebuilds it, so `stats is self.stats` is a sufficient "nothing changed" test. That is a separate change from this one, and I did not want to smuggle it in here.

## Two small things noticed along the way

`_get_ttl_hash()` ignores its `ttl` argument and returns `datetime.now().second`, so `ttl=30` behaves like `ttl=1` and the value cycles every minute. It is unreachable today precisely because the caches never hit, so it is not worth a PR on its own — flagging it in case the cache is ever sized to be effective.

In `_check_decorator`, `self.stats == self.get_init_value` compares the stats against the bound method rather than its result, so that branch is always `False`. Harmless, because `refresh_timer` starts expired and the first update happens anyway.

Closing this — the approach is wrong. Happy to open a focused PR for the `processlist` / `programlist` waste if that direction looks right to you.

PR #3648, opened 2026-08-05
fix: keep zero-valued fields when aggregating processes into programs

Dipet:
#### Description

`update_program_dict()` merges the per-process `cpu_times` and `memory_info` dicts with `Counter` arithmetic:

```python
program['cpu_times'] = dict(Counter(program['cpu_times'] or {}) + Counter(p['cpu_times'] or {}))
```

`Counter.__add__` drops every key whose sum is not positive. `iowait`, `children_user` and `shared` are legitimately `0.0` for most processes, so those keys silently disappear from the aggregated program:

```python
>>> dict(Counter({'user': 1.0, 'system': 0.0, 'iowait': 0.0}) + Counter({'user': 2.0, 'system': 0.0, 'iowait': 0.0}))
{'user': 3.0}
```

On a live host every process has the same five `cpu_times` keys, while the programs built from them end up with **six different key sets**, down to a single `{'system'}`. **78 of 1453 programs** carried an incomplete `cpu_times` dict, among them `node`, `python` and `MainThread`. Anything reading `program['cpu_times']['iowait']` raises `KeyError`, and the REST API and exporters emit a schema that varies from program to program.

Replaced with plain dict addition, which keeps zeros. It is also considerably cheaper: the old form built three `Counter` objects per field per process, 87 000 allocations for 15 900 processes.

Measured on 15 921 processes, the two implementations run back to back in one process:

| `processes_to_programs()` | median |
|---|---|
| develop | 144.3 ms |
| this PR | 53.0 ms |

**2.72x**, and programs with an incomplete `cpu_times` go from 78 to 0.

#### Tests

Adds `tests/test_programs_aggregation.py`: zero totals keep their field, missing fields are added, `None` operands are tolerated, and an aggregated program keeps the full `cpu_times` and `memory_info` key sets.

Full suite on Linux (web tests skipped, no FastAPI/selenium in this environment): **638 passed, 4 failed**. The four — `test_perf_update`, `test_memoryleak_no_history` and two in `test_actions_sanitize` — fail identically on unmodified `develop` on this machine. `test_perf_update` asserts more than one full update per second, which this host (15 900 processes) cannot manage either way.

#### Resume

* Bug fix: yes
* New feature: no
* Fixed tickets:

### DIFF
--- a/glances/programs.py
+++ b/glances/programs.py
@@ -6,8 +6,6 @@
 # SPDX-License-Identifier: LGPL-3.0-only
 #
 
-from collections import Counter
-
 # from glances.logger import logger
 
 # This constant defines the list of available processes sort key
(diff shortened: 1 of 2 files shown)


nicolargo:
@Dipet Thanks for the refactoring ! 

PR #3650, opened 2026-08-05
Remove `pyinstrument` from runtime dependencies

C0rn3j:
#### Description

Has some superficial corrections and also a non-superficial removal of `pyinstrument` from all package groups except for `dev`.

This was causing a `pip check` fail with the Arch Linux package - it seems that `pyinstrument` is part of profilers, so I assumed anywhere where `memray` isn't, `pyinstrument` shouldn't be either.

By the way, what's the point of duplicating the package groups?

You already use `pyproject.toml`, using runtime dependencies, optional dependencies and the internal dependency groups - what's the reason to also define `all-requirements.txt`, `dev-requirements.txt`, `docker-requirements.txt` (notably missing from `pyproject.toml`) and `requirements.txt`?

#### Resume

* Bug fix: yes
* New feature: no
* Fixed tickets: N/A

### DIFF
--- a/NEWS.rst
+++ b/NEWS.rst
@@ -377,7 +377,7 @@ Enhancements:
 
 Bugs corrected:
 
-* [MACOS] Glances not showing Processes on MacOS #3100
+* [MACOS] Glances not showing Processes on macOS #3100
 * Last dev build broke Homepage API calls ? only 1 widget still working #3322
 * Cloud plugin always generate communication with 169.254.169.254, even if the plugin is disabled #3316
 * API response delay (3+ minutes) when VMs are running #3317
@@ -1029,7 +1029,7 @@ Bugs corrected:
 
 * Correct issue (error message) concerning the Cloud plugin - Related to #2392
 * InfluxDB2 export doesn't process folders correctly - missing key #2327
-* Index error when displaying programs on MacOS #2360
+* Index error when displaying programs on macOS #2360
 * Dissociate 2 sensors with exactly the same names #2280
 * All times displayed in UTC - Container not using TZ/localtime (Docker) #2278
 * It is not possible to return API data for a particular mount point (FS plugin) #1162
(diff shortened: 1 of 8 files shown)


nicolargo:
Hi @C0rn3j thanks for the PR. I will have a look asap.

Requirements files are auto-generated by the 'make requirements' target and used to build the Docker images. Last time i have a look, i did not success to use pyproject file instead.

nicolargo:
@C0rn3j  Thanks for the PR ! Just merged into develop branch.

C0rn3j:
> Last time i have a look, i did not success to use pyproject file instead.

Should work just fine nowadays via `pip install ".[dev]"` for example, especially since you seem to be using latest Python - not sure about the snap packaging capabilities though, the docker.txt definitions for might have to stay for snap.

By the way, I glanced at the Dockerfile, doing the following is no bueno:

```Dockerfile
# Upgrade the system
RUN apt-get update \
  && apt-get upgrade -y

# Install build-time dependencies
RUN apt-get install -y --no-install-recommends \
    python3-dev \
    python3-venv \
    python3-pip \
    python3-wheel \
    libzmq5 \
    musl-dev \
    build-essential

RUN apt-get clean \
  && rm -rf /var/lib/apt/lists/*
```

Each RUN command creates a layer, so you're creating a layer which has the update lists and files, another layer which has extra deps, and then finally another layer which deletes the lists and package cache - but you are never going to see the size savings for that, since they already exist in the previous layer which has to get downloaded - so what you're doing is actually *increasing* the size taken, as the layer is effectively pointless in that state.

Doing it all in one RUN command fixes it:

```Dockerfile
RUN <<EOR
# Upgrade the system
apt-get update
apt-get upgrade -y

# Install build-time dependencies
apt-get install -y --no-install-recommends \
    python3-dev \
    python3-venv \
    python3-pip \
    python3-wheel \
    libzmq5 \
    musl-dev \
    build-essential

apt-get clean
rm -rf /var/lib/apt/lists/*
EOR
```

PR #3654, opened 2026-08-07
perf: build the process views on first read instead of on every refresh

Dipet:
#### Description

`update_views()` builds a view dict for every field of every item. For `processlist` that is
every process on the machine, while the UI shows a few dozen rows — and `msg_curse()` never
reads those views, it builds its own decorations inline. They only leave the plugin through
`getViewsProcesslist`.

On a host with ~16 000 processes that is 255 744 calls to `_build_view_for_field` per refresh.
The work is flat: no hot spot inside, the cost is the count.

`processlist` now marks its views as out of date instead of rebuilding them, and `get_views()`
calls the inherited `update_views()` the first time somebody asks. `set_views()` and
`reset_views()` clear the mark, so a reset stays reset. Nothing outside the plugin reads
`.views` directly and every reader goes through `get_views()`, so consumers still see a plain,
complete dict — the REST API included.

| `processlist.update_views()` on 16 000 processes | median |
|---|---|
| develop | 216.1 ms |
| this branch | 0.0 ms |

The first `get_views()` after a refresh costs what `update_views()` used to and the result is
reused until the next one. In curses mode nobody calls it. The full `stats.update()` tick goes
from 0.635 s to 0.184 s on the same machine.

It scales with the process count on any box; a synthetic list is enough to see it:

```
 processes   update_views  views built  rows on screen
       100          0.4ms          100              50
     10000         51.6ms        10000              50
     20000        113.8ms        20000              50
```

Sorting and the process table are untouched: they work off `stats`, not `views`.

An earlier version of this put a `lazy_views` flag in `GlancesPluginModel`. Keeping it in the
one plugin that wants it turned out smaller and avoids a trap for the others: the base class
would have had to detect subclasses that decorate `self.views` right after calling `super()`,
which is what 19 plugins do.

#### Tests

`tests/test_lazy_views.py`. `test_refresh_builds_nothing` fails on `develop`; the rest were
checked by breaking the implementation on purpose — removing the build from `get_views()` fails
two of them, dropping the flag reset in `reset_views()` fails another. Run together with
`tests/test_core.py`: 54 passed.

#### Resume

* Bug fix: no (performance)
* New feature: no
* Fixed tickets:

### DIFF
--- a/glances/plugins/plugin/model.py
+++ b/glances/plugins/plugin/model.py
@@ -56,10 +56,6 @@ fields_unit_type = {
 class GlancesPluginModel:
     """Main class for Glances plugin model."""
 
-    # Defer building the views until something reads them. Worth it for a plugin holding far
-    # more items than the UI draws, whose own rendering does not consult them.
-    lazy_views = False
-
     def __init__(self, args=None, config=None, items_history_list=None, stats_init_value={}, fields_description=None):
         """Init the plugin of plugins model class.
 
@@ -121,8 +117,6 @@ class GlancesPluginModel:
 
         # Init the views
         self.views = {}
-        # Set by update_views() when the build is deferred; consumed by get_views()
-        self._views_source = None
 
         # Hide stats if all the hide_zero_fields has never been != 0
         # Default is False, always display stats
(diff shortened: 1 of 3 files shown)


nicolargo:
Hi @Dipet 

just review and merge this PR on develop branch.

For the record, Glances 4.5.x will be the latest versions of the Glances 4.

Glances 5 is under development in the develop-v5 branch. When code will be freeze, you will welcome to contribute for perf PR in this new version.

Thanks !

PR #3655, opened 2026-08-07
perf: reuse the program aggregation until the process list changes

Dipet:
#### Description

`get_list(as_programs=True)` re-runs `processes_to_programs()` on every call, although the
process list it aggregates is only rebuilt by `update()`. The `programlist` plugin asks for it
on every refresh, so the whole list is walked again for a result that cannot have changed.

The aggregation is now kept until the list it was built from is replaced:

| `get_list(as_programs=True)` on ~16 000 processes | median |
|---|---|
| first call after a refresh | 207.8 ms |
| repeated call | 0.00 ms |

`programlist.update()` drops from 153 ms to nothing on the refreshes in between, which on this
machine is most of them.

Replacing the list is what normally invalidates the cache, so `update()` needs no change. A
re-sort is the exception: one branch of `sort_stats()` reorders in place when the primary sort
raises, leaving the same object with a different order, so `get_list(sorted=True)` clears the
cache explicitly.

#### Tests

`tests/test_program_aggregation.py`. `test_aggregation_is_reused_while_the_list_is_unchanged`
fails on `develop`; the others were checked by breaking the implementation on purpose —
removing the explicit invalidation fails the in-place sort test, and neutering the identity
check fails two more. Run together with `tests/test_core.py`: 53 passed.

#### Resume

* Bug fix: no (performance)
* New feature: no
* Fixed tickets:

### DIFF
--- a/glances/processes.py
+++ b/glances/processes.py
@@ -74,7 +74,7 @@ class GlancesProcesses:
         self.processlist = []
         self.reset_processcount()
 
-        # Programs aggregated from processlist, kept until the list itself is replaced
+        # Aggregation of processlist, valid while _programs_source is still the current list
         self._programs = []
         self._programs_source = None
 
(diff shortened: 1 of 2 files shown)


nicolargo:
Thanks again @Dipet 

Just merged into develop !

PR #3661, opened 2026-08-10
fix(network): use the TX alert for bytes_sent_rate_per_sec decoration

### LINKED ISSUE #3660: Web UI colours network TX rate with the RX alert (bytes_sent_rate_per_sec gets alert_rx)
### Description

In `glances/plugins/network/__init__.py`, the decoration (colour/alert state) for
`bytes_sent_rate_per_sec` is assigned the **RX** alert instead of the TX alert:

https://github.com/nicolargo/glances/blob/master/glances/plugins/network/__init__.py#L233-L236

```python
self.views[i[self.get_key()]]['bytes_recv']['decoration'] = alert_rx
self.views[i[self.get_key()]]['bytes_recv_rate_per_sec']['decoration'] = alert_rx
self.views[i[self.get_key()]]['bytes_sent']['decoration'] = alert_tx
self.views[i[self.get_key()]]['bytes_sent_rate_per_sec']['decoration'] = alert_rx   # <-- alert_tx?
```

Line 235 correctly gives `bytes_sent` the `alert_tx` value; line 236 gives the
`_rate_per_sec` variant `alert_rx`.

This matters because the **Web UI reads the `_rate_per_sec` key**, while the curses
UI reads the non-rate one:

https://github.com/nicolargo/glances/blob/master/glances/outputs/static/js/components/plugin-network.vue#L30

```html
:class="getDecoration(network.interfaceName, 'bytes_sent_rate_per_sec')"
```

So the two front ends can show different alert states for the same interface in the
same poll.

### Steps to reproduce

1. Default `[network]` limits (`rx_careful=70 … rx_critical=90`, same for tx).
2. An interface with a known `speed` (so `get_alert` has a `maximum`) — e.g. a
   1 Gbps `eth0`.
3. Drive asymmetric traffic: high RX, low TX (~950 Mbps down, ~10 Mbps up).
4. Compare `glances` (curses) with the Web UI at the same moment.

### Expected

TX rate is ~1% of link speed, so it should be undecorated in both UIs.

### Actual

- Curses: `Rx/s` red (correct), `Tx/s` undecorated (correct).
- Web UI: `Tx/s` shown in the **critical** colour, because it inherited the RX alert.

The inverse also occurs — a saturated uplink with a quiet downlink shows TX
undecorated in the Web UI.

### Suggested fix

```diff
-            self.views[i[self.get_key()]]['bytes_sent_rate_per_sec']['decoration'] = alert_rx
+            self.views[i[self.get_key()]]['bytes_sent_rate_per_sec']['decoration'] = alert_tx
```

### Environment

- Glances `4.5.7_dev01`, `master` @ `fed67b21a8f7e1ccb41823419791683ed38aeeaa`
- Found by reading the source; I have not run a live reproduction, so please treat
  the reproduction steps as derived from the code path rather than observed.

### How this was found

Part of an audit looking specifically for cases where a displayed value and the
severity attached to it are computed from different quantities. Happy to open a PR
for the one-line change if that's useful.

siaklin99-gif:
Fixes #3660.

`bytes_sent_rate_per_sec` was being decorated with `alert_rx` instead of `alert_tx`. The line directly above it already does the right thing for `bytes_sent`:

```python
self.views[...]['bytes_sent']['decoration'] = alert_tx
self.views[...]['bytes_sent_rate_per_sec']['decoration'] = alert_rx   # <-- this one
```

It's user-visible because the two front ends read different keys — the curses UI reads `bytes_sent`, the Web UI reads `bytes_sent_rate_per_sec` ([plugin-network.vue#L30](https://github.com/nicolargo/glances/blob/develop/glances/outputs/static/js/components/plugin-network.vue#L30)). On an interface with asymmetric traffic they can therefore disagree in the same poll: a 1 Gbps link at ~950 Mbps RX / ~10 Mbps TX shows `Tx/s` undecorated in the terminal and in the **critical** colour in the browser.

One-line change, against `develop` per CONTRIBUTING. I found this by reading the source rather than from a live reproduction, so a second pair of eyes on the expected behaviour is welcome — but the asymmetry with line 235 looks unambiguous.

### DIFF
--- a/glances/plugins/network/__init__.py
+++ b/glances/plugins/network/__init__.py
@@ -233,7 +233,7 @@ class NetworkPlugin(GlancesPluginModel):
             self.views[i[self.get_key()]]['bytes_recv']['decoration'] = alert_rx
             self.views[i[self.get_key()]]['bytes_recv_rate_per_sec']['decoration'] = alert_rx
             self.views[i[self.get_key()]]['bytes_sent']['decoration'] = alert_tx
-            self.views[i[self.get_key()]]['bytes_sent_rate_per_sec']['decoration'] = alert_rx
+            self.views[i[self.get_key()]]['bytes_sent_rate_per_sec']['decoration'] = alert_tx
 
     def _msg_curse_header(self, args, name_max_width):
         """Return the header curse lines."""


nicolargo:
Thanks for the issue report and the patch  @siaklin99-gif  !

Issue #3456, opened 2026-02-25
Glances won't start - Permission denied: '/usr/share/libdrm/amdgpu.ids'


YamiYukiSenpai:
**Check the bug**
Before filling this bug report, please search if a similar issue already exists.
In this case, just add a comment on this existing issue.

**Describe the bug**
Glances won't start

**To Reproduce**
1. Start Glances with the following options '...'
2. Press the key '....'
3. See error

**Expected behavior**
The web version of glances runs.

I only noticed it because the Snap refreshed on my desktop.  My server still has 4.4.1 from November so it wasn't refreshed

**Screenshots**
N/A

**Environement (please complete the following information)**
 - Operating System (lsb_release -a or OS name/version):
```
> lsb_release -a
LSB Version:    n/a
Distributor ID: Garuda
Description:    Garuda Linux Broadwing
Release:        Soaring
Codename:       Broadwing
```
 - Glances & psutil versions:
```
> glances -V
Glances version:        4.5.1_dev03
Glances API version:    4
PsUtil version:         7.1.1
Log file:               /tmp/glances-yamiyuki.log
```
 - How do you install Glances (Pypi package, script, package manager, source): Snap
 - Glances test:
```
> glances --issue
Traceback (most recent call last):
  File "/snap/glances/2660/bin/glances", line 6, in <module>
    sys.exit(main())
             ^^^^^^
  File "/snap/glances/2660/lib/python3.12/site-packages/glances/__init__.py", line 191, in main
    start(config=core.get_config(), args=core.get_args())
  File "/snap/glances/2660/lib/python3.12/site-packages/glances/__init__.py", line 150, in start
    mode = GlancesMode(config=config, args=args)
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/snap/glances/2660/lib/python3.12/site-packages/glances/standalone.py", line 79, in __init__
    self.stats.update()
  File "/snap/glances/2660/lib/python3.12/site-packages/glances/stats.py", line 287, in update
    self.update_plugin(p)
  File "/snap/glances/2660/lib/python3.12/site-packages/glances/globals.py", line 575, in inner
    return _func(weakref.ref(self), *args, ttl_hash=_get_ttl_hash(ttl), **kwargs)
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
[...]

PR #3664, opened 2026-08-13
Fix AMD GPU name resolution when the card is only listed in AMD's `amdgpu.ids`

### LINKED ISSUE #3663: AMD GPU name falls back to "AMD GPU" when the card is only in AMD's `amdgpu.ids`
**Check the bug**

We've searched the open and closed issues and found nothing covering the `amdgpu.ids` lookup path.

**Describe the bug**

`get_device_name()` in `glances/plugins/gpu/cards/amd.py` resolves the card name by matching its PCI device/revision ID against a lookup table, and the path to that table is hard-coded to a single location:

```python
AMDGPU_IDS_FILE: str = '/usr/share/libdrm/amdgpu.ids'
```

There are commonly two copies of that file on an AMD host, shipped by two different packages:

| Path | Package | Origin |
| --- | --- | --- |
| `/usr/share/libdrm/amdgpu.ids` | `libdrm-common` | distribution, frozen at release |
| `/opt/amdgpu/share/libdrm/amdgpu.ids` | `libdrm-amdgpu-common` | AMD, via `amdgpu-install` / ROCm |

Glances only ever reads the first. We saw this on two separate machines:

| Host | GPUs | Distribution copy | AMD copy |
| --- | --- | --- | --- |
| host A, Ubuntu 22.04.5 | 8 × Instinct MI250X (`740C:01`) | 2022-08-31, 313 entries | 2026-01-20, 347 entries |
| host B, Ubuntu 22.04.2 | 8 × Instinct MI355X (`75A3:00`) | 2022-08-31, 313 entries | 2026-01-09, 348 entries |

When the card is missing from the distribution copy the lookup misses and the function returns the literal fallback:

```python
return 'AMD GPU'
```

So every card is labelled `AMD GPU`. Utilisation, memory and temperature are all correct, so this is purely a naming defect. Present since the name lookup was added in ec11dff2, first released in v4.5.0. Host B reproduces it on the released v4.5.5 from PyPI and host A on `develop` (fed67b21), so it is not specific to a source checkout.

**To Reproduce**

The trigger is a card that the distribution's `libdrm` predates. On both hosts above the distribution list is Ubuntu 22.04's, frozen in 2022, and 87 of the IDs in AMD's copy are absent from it. Those 87 cover most of what people actually run ROCm or a recent driver on:

- **Instinct**, 21 IDs: MI100, MI210, MI250/MI250X, MI300A/MI300X/MI308X, MI325X, MI350X, MI355X
- **Radeon RX 7000 (RDNA 3)**, 15 IDs: RX 7900 XTX/XT/GRE, 7800 XT, 7700 XT, 7600 XT and the mobile parts
- **Radeon RX 9000 (RDNA 4)**, 5 IDs: RX 9070 XT/GRE, 9060 XT
- **Radeon PRO**, 16 IDs: W7900/W7800/W7700/W7600/W7500, V520/V620/V710, AI PRO R9700

Steps:

1. Confirm the two files disagree about your card (`75A3` here, the MI355X):

```
$ grep -cE '^[0-9A-Fa-f]{4},' /usr/share/libdrm/amdgpu.ids /opt/amdgpu/share/libdrm/amdgpu.ids
/usr/share/libdrm/amdgpu.ids:313
/opt/amdgpu/share/libdrm/amdgpu.ids:348

$ grep '^75A3' /usr/share/libdrm/amdgpu.ids
$ grep '^75A3' /opt/amdgpu/share/libdrm/amdgpu.ids
75A3,	00,	AMD Instinct MI355X
```

2. Start `glances` and look at the GPU panel. Every card reads `AMD GPU`.

Hardware-independent check, straight against the backend:

```python
from glances.plugins.gpu.cards.amd import AmdGPU
print({g['name'] for g in AmdGPU().get_device_stats()})
# {'AMD GPU'}   <-- expected {'Instinct MI355X'}
```

**Expected behavior**

The model name, as it appears in AMD's own table:

```
8 Instinct MI355X
Instinct    0% mem  16%
```

**Screenshots**

Image

**Environement (please complete the following information)**

Host A, 8 × Instinct MI250X:

- Operating System: Ubuntu 22.04.5 LTS (jammy), kernel 5.15.0-171-generic
- Glances & psutil versions:

```
Glances version:	4.5.7_dev01
Glances API version:	4
PsUtil version:		7.2.2
```

- How do you install Glances: from source, `uv tool install -e`
- Glances test (`glances --issue`, GPU section):

```
gpu           [OK]    0.00217s key=gpu_id
[{'fan_speed': None, 'gpu_id': 'amd0', 'key': 'gpu_id', 'mem': 0, 'name': 'AMD GPU', 'proc': 0, 'temperature': None}, ...]
```

Host B, 8 × Instinct MI355X:

- Operating System: Ubuntu 22.04.2 LTS (jammy), kernel 5.15.0-70-generic
- Glances & psutil versions:

```
Glances version:	4.5.5
Glances API version:	4
PsUtil version:		7.2.2
```

- How do you install Glances: released version from PyPI, `uv tool install glances`
- Glances test (`glances --issue`, GPU section):

```
gpu           [OK]    0.00219s key=gpu_id
[{'fan_speed': None, 'gpu_id': 'amd0', 'key': 'gpu_id', 'mem': 0, 'name': 'AMD GPU', 'proc': 0, 'temperature': 34}, ...]
```

**Additional context**

Both hosts carry `libdrm-amdgpu-common` alongside the distribution `libdrm-common`, which is the normal result of running `amdgpu-install` or installing ROCm.

williamqwu:
#### Description

Fixes https://github.com/nicolargo/glances/issues/3663, where `get_device_name()` in `glances/plugins/gpu/cards/amd.py` hard-codes its lookup table to `/usr/share/libdrm/amdgpu.ids`. That copy is frozen at distribution release, so any newer card misses the lookup and falls through to the literal `'AMD GPU'`. Hosts running ROCm or `amdgpu-install` also carry `libdrm-amdgpu-common`, which ships a current copy of the same table at `/opt/amdgpu/share/libdrm/amdgpu.ids`; Glances simply never looked there.

This PR makes the constant an ordered tuple and walks it until a card matches. Two details carry the safety of the change. **The loop stops on a match, not on a readable file:** 51 device IDs live in the distribution list and are absent from AMD's, 50 of them Raven Ridge and Picasso APUs, so returning at the first file that merely exists would have regressed all of them to `AMD GPU`. **The distribution list is searched first,** which makes the change strictly additive: a name that already resolved keeps resolving to the exact same string, and the patch can only ever turn `AMD GPU` into a real name.

I did try AMD's list first and measured it before dropping it. Of the 259 IDs present in both files, 90 render differently after the `AMD ` / ` Graphics` stripping the function applies: 63 are cosmetic `(TM)` insertions and 27 are substantive rewrites (`Radeon Vega 3` becomes `Ryzen Embedded R1305G with Radeon Vega Gfx`). That is user-visible churn on machines that were never broken, and `(TM)` costs five characters in a field the header truncates at 17.

Before:

```
8 AMD GPU
AMD GPU   0% mem  16%    # eight cards, none of them named
```

After:

```
8 Instinct MI355X
Instinct    0% mem  16%  # name from AMD's table
```

Verified on two hosts, 8 × MI250X and 8 × MI355X, and against both real databases comparing old behaviour with new: `740C:01`, `75A3:00` and `744C:C8` go from `AMD GPU` to their model names, while the APU IDs `15D8:CF` and `15DD:C1`, `6600:81`, an unknown `FFFF:FF`, and any host without `/opt` are all byte-identical to before. NVIDIA is untouched by construction, since those names come from NVML in `cards/nvidia.py` and never reach `amdgpu.ids`, and the AMD backend only admits DRM devices exposing `gpu_busy_percent`. Snap strict confinement needs no new handling: `read_file()` already wraps `open()` in `try/except`, so a blocked `/opt` returns `None` and the loop moves on, the same guard added for #3456. Cost is nil when the distribution list matches, since the loop returns before touching `/opt`; on a miss it is one extra read and one regex scan, once per card per process because `get_device_name()` is `@functools.cache`d.

`pytest tests/test_plugin_gpu.py` passes (41), `ruff check` and `ruff format --check` are clean. Note that `AMDGPU_IDS_FILE: str` is renamed and retyped to `AMDGPU_IDS_FILES: tuple[str, ...]`; the only two references in the tree are the definition and its use in the same function. Happy to add a unit test for the multi-path lookup and a `NEWS.rst` entry if you want them here, as there is currently no test covering AMD's `get_device_name()`.

#### Resume

- Bug fix: yes, fixes https://github.com/nicolargo/glances/issues/3663
- New feature: no

| :warning: WARNING           |
|:----------------------------|
| Warning: constant `AMDGPU_IDS_FILES` is changed from type `str` to `tuple`.     |

### DIFF
--- a/glances/plugins/gpu/cards/amd.py
+++ b/glances/plugins/gpu/cards/amd.py
@@ -46,7 +46,10 @@ import re
 
 DRM_ROOT_FOLDER: str = '/sys/class/drm'
 DEVICE_FOLDER_PATTERN: str = 'card[0-9]*/device'
-AMDGPU_IDS_FILE: str = '/usr/share/libdrm/amdgpu.ids'
+# Looked up in order, first match wins. The distribution list comes first so that an already
+# resolved name never changes; the list shipped by AMD (amdgpu-install / ROCm) is only consulted
+# for the cards the distribution does not know, typically the Instinct data center ones.
+AMDGPU_IDS_FILES: tuple[str, ...] = ('/usr/share/libdrm/amdgpu.ids', '/opt/amdgpu/share/libdrm/amdgpu.ids')
 PCI_DEVICE_ID: str = 'device'
 PCI_REVISION_ID: str = 'revision'
 GPU_PROC_PERCENT: str = 'gpu_busy_percent'
@@ -125,16 +128,17 @@ def get_device_name(device_folder: str) -> str:
     # Table source: https://cgit.freedesktop.org/drm/libdrm/tree/data/amdgpu.ids
     device_id = read_file(device_folder, PCI_DEVICE_ID)
     revision_id = read_file(device_folder, PCI_REVISION_ID)
-    amdgpu_ids = read_file(AMDGPU_IDS_FILE)
-    if device_id and revision_id and amdgpu_ids:
+    if device_id and revision_id:
         # Strip leading "0x" and convert to uppercase hexadecimal
         device_id = device_id[2:].upper()
         revision_id = revision_id[2:].upper()
         # Syntax:
         # device_id,	revision_id,	product_name        <-- single tab after comma
         pattern = re.compile(f'^{device_id},\\s{revision_id},\\s(?P<product_name>.+)$', re.MULTILINE)
-        if match := pattern.search(amdgpu_ids):
-            return match.group('product_name').removeprefix('AMD ').removesuffix(' Graphics')
+        for amdgpu_ids_file in AMDGPU_IDS_FILES:
+            amdgpu_ids = read_file(amdgpu_ids_file)
+            if amdgpu_ids and (match := pattern.search(amdgpu_ids)):
+                return match.group('product_name').removeprefix('AMD ').removesuffix(' Graphics')
 
     return 'AMD GPU'
 


nicolargo:
Thanks for the issue report and PR @williamqwu  !

PR #3665, opened 2026-08-15
fix: compare stats to the init value, not to the bound method

VXNCXNX:
## What's broken

In `_check_decorator`, `glances/plugins/plugin/model.py:1278`:

```python
if self.is_enabled() and (self.refresh_timer.finished() or self.stats == self.get_init_value):
```

`self.get_init_value` is the bound method, not its result, so `self.stats` never equals it and that half of the condition is dead.

## Why it matters

The branch exists for the fetch-error path. On an SNMP error, `cpu/__init__.py:249` and `:271`, and `mem/__init__.py:227` and `:245`, call `self.reset()`, which sets `self.stats = self.get_init_value()`. The intent is that the next poll retries straight away instead of serving empty stats until the refresh interval expires. With the missing parentheses that retry never happens.

## The fix

Call it. One character.

I grepped for the same mistake elsewhere. The only other `== self.<attr>` comparisons are `glances_curses_browser.py:77` and `glances_restful_api.py:99`, both against plain attributes, so neither is affected.

## Verification

`tests/test_plugin_model.py` applies the decorator to a fake plugin and asserts update runs when stats are at the init value with the timer unfinished, and does not run when stats are populated. The first fails without the fix.

One thing to flag, because it is the only part of this that is not mechanical. Three tests in `tests/test_plugin_fs.py` started failing with `KeyError('/')` once the fix landed. They were passing vacuously: they call `update()` and then `msg_curse()`, and with the dead branch `update()` was a no-op after a preceding test's `reset()`, so `msg_curse` iterated an empty stats list and asserted nothing. With the fix, `update()` refetches real filesystems while `self.views` is still empty.

They now call `update_views()` after `update()`, which is the order production uses in `stats.py:319-320`. That felt like the right call since the tests were out of step with the real call sequence, but if you would rather `msg_curse` defend itself against empty views, say so and I will do that instead.

`pytest tests/` is otherwise unchanged: the two `test_actions_sanitize` failures are the same before and after on my machine.

### DIFF
--- a/glances/plugins/plugin/model.py
+++ b/glances/plugins/plugin/model.py
@@ -1275,7 +1275,7 @@ class GlancesPluginModel:
         """
 
         def wrapper(self, *args, **kw):
-            if self.is_enabled() and (self.refresh_timer.finished() or self.stats == self.get_init_value):
+            if self.is_enabled() and (self.refresh_timer.finished() or self.stats == self.get_init_value()):
                 # Run the method
                 ret = fct(self, *args, **kw)
                 # Reset the timer
--- a/tests/test_plugin_fs.py
+++ b/tests/test_plugin_fs.py
@@ -257,18 +257,21 @@ class TestFsPluginMsgCurse:
     def test_msg_curse_returns_list(self, fs_plugin):
         """Test that msg_curse returns a list."""
         fs_plugin.update()
+        fs_plugin.update_views()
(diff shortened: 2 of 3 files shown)


nicolargo:
Thanks for the PR @VXNCXNX 

Just merged into develop.

PR #3667, opened 2026-08-15
feat(fs): allow free-space display in config

nightcityblade:
#### Description

Allow the `[fs]` section to set `free_space=true`, enabling the same display mode as `--fs-free-space`. The command-line flag remains effective, and the sample configuration plus a focused regression test document the new setting.

Fixes #3659.

Tests:

- `uv run pytest -q tests/test_main.py` (1 passed)
- `uv run pytest -q tests/test_core.py` (48 passed, 4 skipped)
- `uv run ruff check glances/main.py tests/test_main.py`
- `uv run ruff format --check glances/main.py tests/test_main.py`

#### Resume

* Bug fix: no
* New feature: yes
* Fixed tickets: #3659

PR #3670, opened 2026-08-19
fix(containers): aggregate network stats over all container interfaces (#3669)

### LINKED ISSUE #3669: Fix container network accounting: sum all interfaces, not only `eth0`
## Context
 
`DockerStatsFetcher._get_network_stats()` in
`glances/plugins/containers/engines/docker.py` reads network counters from a
single hardcoded interface:
 
```python
eth0_stats = self._streamer.stats.get('networks', {}).get('eth0')
```
 
The Docker stats endpoint returns one entry per container interface. A container
attached to several Docker networks (`docker network connect`, or several
networks declared in a Compose file) gets `eth0`, `eth1`, … and everything past
`eth0` is silently dropped. Reported RX/TX rates for those containers are too
low.
 
Verified on both `develop` and `develop-v5`: the file is **byte-identical** on
the two branches, so the same patch applies to both.
 
## Scope
 
**In scope:** `_get_network_stats()` in
`glances/plugins/containers/engines/docker.py`, plus unit tests.
 
**Out of scope — do not touch:**
 
- `glances/plugins/containers/engines/podman.py`. Podman exposes only a
  pre-aggregated `NetIO` string; there is no per-interface data to sum and no
  bug to fix there.
- The containers plugin model, the TUI, or the WebUI. The output contract of
  `_get_network_stats()` is unchanged: same keys, same types, same units.
- Any adjacent code, comment or formatting in `docker.py` (CLAUDE.md §3).
## Non-regression requirements
 
These must all hold after the change:
 
1. Single-interface containers (the majority) produce **identical values** to
   the current implementation.
2. Containers started with `--network host` have no `networks` key in the stats
   payload. The method must keep returning `None`, exactly as today.
3. The returned dict keeps the same shape: `cumulative_rx`, `cumulative_tx`, and
   — only when previous stats exist — `time_since_update`, `rx`, `tx`.
4. No new dependency, no new configuration key.
## Plan
 
```md
1. Patch `_get_network_stats()` to aggregate across interfaces
   → verify: `make lint && make format` clean; the diff touches only the
     interface-selection and summation lines of that one method
 
2. Add `tests/test_plugin_containers.py` covering single-interface,
   multi-interface, host-network, malformed-payload and rate-computation cases
   → verify: `uv run pytest tests/test_plugin_containers.py` passes
 
3. Run the existing suite to confirm no collateral damage
   → verify: `make test` shows no new failures versus the pre-change baseline
 
4. Add a NEWS.rst entry
   → verify: entry follows the existing `.rst` bullet format under the
     unreleased section
```
 
## Step 1 — the patch
 
Replace the interface selection and the `stats` dict construction. Keep the rest
of the method (rate computation from `_old_computed_stats`) untouched.
 
```python
        interfaces_stats = self._streamer.stats.get('networks') or {}
 
        # Checks for net_stats & mandatory fields
        # Note: a container can be attached to several networks (eth0, eth1...)
        # so all the interfaces should be aggregated (issue #XXXX)
        valid_interfaces = [
            interface_stats
            for interface_name, interface_stats in interfaces_stats.items()
            if interface_name != 'lo' and all(field in interface_stats for field in ['rx_bytes', 'tx_bytes'])
        ]
        if not valid_interfaces:
            self._log_debug("Missing Network usage fields")
            return None
 
        # Read the rx/tx stats (in bytes), summed over all the container interfaces
        stats = {
            'cumulative_rx': sum(interface_stats['rx_bytes'] for interface_stats in valid_interfaces),
            'cumulative_tx': sum(interface_stats['tx_bytes'] for interface_stats in valid_interfaces),
        }
```
 
Replace `#XXXX` with the real issue number before committing.
 
Two deliberate choices, do not silently change them:
 
- **`lo` is excluded.** The Docker daemon does not normally report loopback in
  `networks`, but Docker-API-compatible runtimes (nerdctl/containerd) have been
  observed to. Loopback traffic would massively inflate the totals, so the guard
  is cheap insurance rather than defensive noise.
- **Interfaces with missing counters are skipped, not fatal.** Previously a
  malformed `eth0` aborted the whole method. Now one malformed interface among
  several no longer discards the valid ones. `None` is still returned when
  nothing usable remains, which preserves requirement 2 above.
## Decision required before implementing — do not decide alone
 
Summing across interfaces introduces a case that did not exist before: if a
container is attached to or detached from a network **between two samples**, the
interface set changes and `cumulative_rx`/`cumulative_tx` jump or drop. A drop
produces a **negative** `rx`/`tx` delta for one cycle.
 
Note that negative deltas are already possible today on container restart
(counters reset), so this is an aggravation of an existing behaviour rather than
a new class of bug.
 
Options:
 
- **(a) Leave as is.** Strictly surgical; one bogus sample in a rare scenario.
- **(b) Clamp to zero:** `max(0, stats['cumulative_rx'] - old[...])`. Hides the
  glitch, but also hides genuine counter resets and changes existing behaviour
  for the restart case.
**Default: (a).** Do not implement (b) without explicit approval.
 
## Step 2 — tests
 
Create `tests/test_plugin_containers.py`.
 
**Style note:** `tests/test_plugin_sensors.py` is written in pytest style
(fixtures, plain classes, bare `assert`) while older files such as
`tests/test_core.py` use `unittest.TestCase`. Match the sibling
`test_plugin_*.py` convention, i.e. pytest style, per CLAUDE.md §3.
 
`DockerStatsFetcher.__init__` opens a live stats stream, so do not call it.
Build the instance without the constructor and inject the two attributes the
method reads:
 
```python
from types import SimpleNamespace
 
from glances.plugins.containers.engines.docker import DockerStatsFetcher
 
 
def build_fetcher(networks, old_stats=None):
    fetcher = DockerStatsFetcher.__new__(DockerStatsFetcher)
    fetcher._container = SimpleNamespace(id='deadbeef')
    fetcher._streamer = SimpleNamespace(
        stats={'networks': networks} if networks is not None else {},
        last_update_time=100.0,
    )
    fetcher._old_computed_stats = old_stats or {}
    fetcher._last_stats_computed_time = 97.0
    return fetcher
```
 
Cases to cover:
 
| Case | Input | Expected |
|---|---|---|
| Single interface | `{'eth0': {'rx_bytes': 100, 'tx_bytes': 200}}` | `cumulative_rx == 100`, `cumulative_tx == 200` (regression guard) |
| Two interfaces | `eth0` 100/200 + `eth1` 50/25 | `cumulative_rx == 150`, `cumulative_tx == 225` (the actual fix) |
| Loopback present | `eth0` 100/200 + `lo` 999/999 | `lo` excluded, `cumulative_rx == 100` |
| Host network | `networks` key absent | returns `None` |
| Empty networks dict | `{}` | returns `None` |
| All interfaces malformed | `{'eth0': {'rx_bytes': 100}}` | returns `None` |
| One malformed among two | `eth0` valid + `eth1` missing `tx_bytes` | `eth1` skipped, `eth0` still counted |
| Rate computation | two interfaces + `_old_computed_stats['network']` with lower cumulative values | `rx`/`tx` are the summed deltas, `time_since_update` present |
| First sample | no `_old_computed_stats` | no `rx`/`tx`/`time_since_update` keys |
 
## Step 3 — regression run
 
Capture a baseline before touching anything:
 
```bash
make test 2>&1 | tail -30   # baseline
# ... apply the change ...
make test 2>&1 | tail -30   # compare
```
 
Only the new test file may add results. Any other delta is a regression, stop
and report it.
 
## Step 4 — changelog
 
Add to `NEWS.rst`, under the unreleased section, bug-fix list:
 
```rst
* Container network stats now aggregate all the container interfaces
  instead of eth0 only (issue #XXXX)
```
 
## Branches
 
Apply to `develop` first. `develop-v5` currently holds an identical copy of the
file, but **re-verify the divergence before assuming a clean cherry-pick** —
confirm the method body still matches, then port.
 
## Report back
 
- Diff of `docker.py` (should be roughly 10 lines).
- Test file, and its pass output.
- Baseline vs post-change `make test` comparison.
- Your recommendation on the decision point above, with reasoning.

ntdatt812:
#### Description

Fixes #3669.

`DockerStatsFetcher._get_network_stats()` read the container network counters from a
single hardcoded interface:

```python
eth0_stats = self._streamer.stats.get('networks', {}).get('eth0')
```

The Docker stats endpoint returns one entry per container interface, so a container
attached to several networks (`docker network connect`, or several networks in a Compose
file) gets `eth0`, `eth1`, … and everything past `eth0` was silently dropped. Reported
RX/TX for those containers was too low.

The method now sums `rx_bytes`/`tx_bytes` across all interfaces. Its output contract is
unchanged: same keys, same types, same units.

Two deliberate choices, as discussed in the issue:

* **`lo` is excluded.** The Docker daemon does not normally report loopback in `networks`,
  but Docker-API-compatible runtimes have been observed to, and loopback traffic would
  massively inflate the totals.
* **Interfaces with missing counters are skipped, not fatal.** Previously a malformed
  `eth0` aborted the whole method; now one malformed interface among several no longer
  discards the valid ones. `None` is still returned when nothing usable remains, so
  `--network host` containers (no `networks` key) behave exactly as before.

#### On the decision point raised in the issue

The issue asked not to decide alone between **(a) leave as is** and **(b) clamp the delta
to zero**, given that attaching/detaching a network between two samples changes the
interface set and can produce one negative `rx`/`tx` delta.

**This PR implements (a)**, for two reasons:

1. Negative deltas are already reachable today on container restart, when the counters
   reset. (b) would silently swallow that pre-existing signal too, which changes behaviour
   for a case that is not in the scope of this issue.
2. (b) hides a real counter discontinuity rather than reporting it. If the glitch turns
   out to matter in practice, clamping is better addressed once for all cumulative
   counters than locally here.

Happy to switch to (b) if you prefer.

#### Tests

New file `tests/test_plugin_containers.py`, 9 cases, pytest style to match the sibling
`test_plugin_*.py` files. `DockerStatsFetcher.__init__` opens a live stats stream, so the
instance is built via `__new__` with the two attributes the method reads injected.

| Case | Expected |
|---|---|
| Single interface | identical values to the previous implementation (regression guard) |
| Two interfaces | counters summed |
| Loopback present | `lo` excluded |
| Host network (no `networks` key) | `None` |
| Empty `networks` dict | `None` |
| All interfaces malformed | `None` |
| One malformed among two | malformed skipped, valid one still counted |
| Rate computation | `rx`/`tx` are the summed deltas, `time_since_update` present |
| First sample | no `rx`/`tx`/`time_since_update` keys |

Verified the tests actually pin the bug — against the unpatched `docker.py`, exactly the
two cases that target it fail and the seven regression guards pass:

```
FAILED tests/test_plugin_containers.py::TestDockerNetworkStatsAggregation::test_two_interfaces_are_summed
        assert 100 == 150
FAILED tests/test_plugin_containers.py::TestDockerNetworkStatsRates::test_rates_use_the_summed_counters
        assert 60 == 110
2 failed, 7 passed
```

With the patch: `9 passed`.

#### Regression run

Full suite before and after, same machine (Windows 11, Python 3.14, 1 running container):

```
before: 57 failed, 571 passed, 106 skipped, 15 errors
after:  58 failed, 579 passed, 106 skipped, 15 errors
```

The 72 pre-existing failures/errors — `test_restful.py`, `test_xmlrpc.py`, `test_mcp.py`,
`test_browser_restful.py`, `test_amp_secure_popen.py`, `test_webui.py`, `test_perf.py`, all
server/WebUI tests that need a chromedriver and a POSIX socket — are **identical line for
line** before and after. The 9 new tests account for the passed count.

The single delta is `test_memoryleak.py::test_memoryleak_no_history`
(`assert memory_leak < 15000`, got 15524). It is not caused by this change:

* The file passes on its own with the patch applied.
* Its measurement is unstable on this machine, and **it exceeds its own threshold on
  unpatched `develop`**. Three runs each, reading the value the test itself logs:

  | run | unpatched `develop` | this branch |
  |---|---:|---:|
  | 1 | 12 391 B | 16 734 B |
  | 2 | 7 948 B | 20 268 B |
  | 3 | **32 480 B** ❌ | 6 718 B |
  | mean | 17 606 B | **14 573 B** |

  The unpatched spread is 7 948 – 32 480 B, a 4× range against a hard-coded 15 000 B
  threshold, and the unpatched maximum is the largest of all six runs.

* Isolating the changed method removes the ambiguity. Driving both implementations
  200 000 times under `tracemalloc` with the same single-interface payload:

  ```
     old (eth0 only): retained after 200,000 calls =  688 bytes  (0.0034 B/call)
    new (aggregated): retained after 200,000 calls =  656 bytes  (0.0033 B/call)
  ```

  Neither retains anything — that residue is tracemalloc's own bookkeeping — and the new
  form retains marginally less. The memory-leak test performs ~13 calls per run, so the
  per-call difference is well under a byte.

Flagging the threshold separately since it is unrelated to this PR: on a machine with any
background load, `test_memoryleak_no_history` looks like it can fail on a clean checkout.
Happy to open a separate issue if that is useful.

`ruff format` and `ruff check` are clean on both touched files. `rstcheck NEWS.rst` reports
the same 8 pre-existing issues before and after, shifted by the one added line.

#### Branches

Applied to `develop`. I re-verified the branch parity the issue asked about rather than
assuming it: `glances/plugins/containers/engines/docker.py` has blob SHA
`c818def749329c9b2d00171c43cb2f2739fac5f4` on both `develop` and `develop-v5`, so the
cherry-pick to `develop-v5` is clean. Say the word and I'll open the companion PR.

#### Resume

* Bug fix: yes
* New feature: no
* Fixed tickets: #3669

### DIFF
--- a/NEWS.rst
+++ b/NEWS.rst
@@ -8,6 +8,7 @@ Version 4.5.6
 
 Bugs corrected:
 
+* Container network stats now aggregate all the container interfaces instead of eth0 only #3669
 * Alert level decided by dict ordering: a failing+slow URL is downgraded to WARNING, an unscanned URL reports CRITICAL #3632
 * GPU plugin duplicates card name and omits N/A for unavailable metrics in multi-GPU #3631
 * GPU plugin duplicates the utilisation value and paints it with the memory colour in multi-GPU #3630
--- a/glances/plugins/containers/engines/docker.py
+++ b/glances/plugins/containers/engines/docker.py
@@ -169,15 +169,25 @@ class DockerStatsFetcher:
             rx: Number of bytes received
             tx: Number of bytes transmitted
         """
-        eth0_stats = self._streamer.stats.get('networks', {}).get('eth0')
+        interfaces_stats = self._streamer.stats.get('networks') or {}
 
         # Checks for net_stats & mandatory fields
-        if not eth0_stats or any(field not in eth0_stats for field in ['rx_bytes', 'tx_bytes']):
+        # Note: a container can be attached to several networks (eth0, eth1...)
+        # so all the interfaces should be aggregated (issue #3669)
+        valid_interfaces = [
+            interface_stats
+            for interface_name, interface_stats in interfaces_stats.items()
+            if interface_name != 'lo' and all(field in interface_stats for field in ['rx_bytes', 'tx_bytes'])
+        ]
+        if not valid_interfaces:
             self._log_debug("Missing Network usage fields")
             return None
 
-        # Read the rx/tx stats (in bytes)
-        stats = {'cumulative_rx': eth0_stats["rx_bytes"], 'cumulative_tx': eth0_stats["tx_bytes"]}
+        # Read the rx/tx stats (in bytes), summed over all the container interfaces
+        stats = {
+            'cumulative_rx': sum(interface_stats['rx_bytes'] for interface_stats in valid_interfaces),
+            'cumulative_tx': sum(interface_stats['tx_bytes'] for interface_stats in valid_interfaces),
+        }
 
         # Using previous stats to calculate rates
         old_network_stats = self._old_computed_stats.get("network")
--- a/tests/test_plugin_containers.py
+++ b/tests/test_plugin_containers.py
@@ -28,6 +28,8 @@ def build_fetcher(networks, old_stats=None):
 
 
 class TestDockerNetworkStatsAggregation:
+    """Test the Docker network stats aggregation across container interfaces."""
+
     def test_single_interface(self):
         """A single interface keeps the values it had before the aggregation change."""
         fetcher = build_fetcher({'eth0': {'rx_bytes': 100, 'tx_bytes': 200}})
@@ -61,14 +63,18 @@ class TestDockerNetworkStatsAggregation:
 
 
 class TestDockerNetworkStatsNoData:
+    """Test the Docker network stats when no usable interface is reported."""
+
     def test_host_network_returns_none(self):
         """--network host containers have no networks key at all."""
         assert build_fetcher(None)._get_network_stats() is None
 
     def test_empty_networks_returns_none(self):
+        """An empty networks dict yields no stats."""
         assert build_fetcher({})._get_network_stats() is None
 
     def test_all_interfaces_malformed_returns_none(self):
+        """Every interface missing a counter yields no stats."""
         assert build_fetcher({'eth0': {'rx_bytes': 100}})._get_network_stats() is None
 
     def test_one_malformed_among_two_keeps_the_valid_one(self):
@@ -85,7 +91,10 @@ class TestDockerNetworkStatsNoData:
 
 
 class TestDockerNetworkStatsRates:
+    """Test the Docker network stats rate computation."""
+
     def test_rates_use_the_summed_counters(self):
+        """Rates are computed from the summed counters, not from a single interface."""
         fetcher = build_fetcher(
             {
                 'eth0': {'rx_bytes': 100, 'tx_bytes': 200},
@@ -99,6 +108,7 @@ class TestDockerNetworkStatsRates:
         assert 'time_since_update' in stats
 
     def test_first_sample_has_no_rate_keys(self):
+        """Without previous stats there is nothing to compute a rate from."""
         fetcher = build_fetcher({'eth0': {'rx_bytes': 100, 'tx_bytes': 200}})
         stats = fetcher._get_network_stats()
         assert 'rx' not in stats
--- a/tests/test_plugin_containers.py
+++ b/tests/test_plugin_containers.py
@@ -1,0 +1,106 @@
+#!/usr/bin/env python
+#
+# Glances - An eye on your system
+#
+# SPDX-FileCopyrightText: 2026 Nicolas Hennion <nicolas@nicolargo.com>
+#
+# SPDX-License-Identifier: LGPL-3.0-only
+#
+
+"""Tests for the Containers plugin engines (issue #3669 network aggregation)."""
+
+from types import SimpleNamespace
+
+from glances.plugins.containers.engines.docker import DockerStatsFetcher
+
+
+def build_fetcher(networks, old_stats=None):
+    """Build a DockerStatsFetcher without opening a live stats stream."""
+    fetcher = DockerStatsFetcher.__new__(DockerStatsFetcher)
+    fetcher._container = SimpleNamespace(id='deadbeef')
+    fetcher._streamer = SimpleNamespace(
+        stats={'networks': networks} if networks is not None else {},
+        last_update_time=100.0,
+    )
+    fetcher._old_computed_stats = old_stats or {}
+    fetcher._last_stats_computed_time = 97.0
+    return fetcher
+
+
+class TestDockerNetworkStatsAggregation:
+    def test_single_interface(self):
+        """A single interface keeps the values it had before the aggregation change."""
+        fetcher = build_fetcher({'eth0': {'rx_bytes': 100, 'tx_bytes': 200}})
+        stats = fetcher._get_network_stats()
+        assert stats['cumulative_rx'] == 100
+        assert stats['cumulative_tx'] == 200
+
+    def test_two_interfaces_are_summed(self):
+        """A container attached to two networks reports the sum of both."""
+        fetcher = build_fetcher(
+            {
+                'eth0': {'rx_bytes': 100, 'tx_bytes': 200},
+                'eth1': {'rx_bytes': 50, 'tx_bytes': 25},
+            }
+        )
+        stats = fetcher._get_network_stats()
+        assert stats['cumulative_rx'] == 150
+        assert stats['cumulative_tx'] == 225
+
+    def test_loopback_is_excluded(self):
+        """Loopback traffic is not counted, even when the runtime reports it."""
+        fetcher = build_fetcher(
+            {
+                'eth0': {'rx_bytes': 100, 'tx_bytes': 200},
+                'lo': {'rx_bytes': 999, 'tx_bytes': 999},
+            }
+        )
+        stats = fetcher._get_network_stats()
+        assert stats['cumulative_rx'] == 100
+        assert stats['cumulative_tx'] == 200
+
+
+class TestDockerNetworkStatsNoData:
+    def test_host_network_returns_none(self):
+        """--network host containers have no networks key at all."""
+        assert build_fetcher(None)._get_network_stats() is None
+
+    def test_empty_networks_returns_none(self):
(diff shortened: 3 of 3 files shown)


ntdatt812:
Two red checks on this PR, both traced — neither is a defect in the change, but you should decide on the second one.

**AppVeyor** — red on every open PR right now, not just this one: #3667, #3665 and #3613 all show `continuous-integration/appveyor/pr = failure` on their current heads. Nothing in this branch touches it.

**Codacy — "13 new issues"** — I reproduced the exact number locally. All 13 are `W0212 protected-access`, and they are the test harness itself:

```
$ pylint --disable=all --enable=W0212 tests/test_plugin_containers.py
tests/test_plugin_containers.py:20:4   W0212: Access to a protected member _container
tests/test_plugin_containers.py:21:4   W0212: Access to a protected member _streamer
tests/test_plugin_containers.py:25:4   W0212: Access to a protected member _old_computed_stats
tests/test_plugin_containers.py:26:4   W0212: Access to a protected member _last_stats_computed_time
tests/test_plugin_containers.py:36:16  W0212: Access to a protected member _get_network_stats
... (9 × _get_network_stats, one per test)
```

4 attribute injections in `build_fetcher` + 9 calls to `_get_network_stats` = 13. That is the harness the issue asked for — `DockerStatsFetcher.__init__` opens a live stats stream, so the instance has to be built via `__new__` with the attributes the method reads injected, and the method under test is private.

`tests/test_core.py` already does the same kind of thing (~22 protected accesses); Codacy only counts what a diff adds, so those don't show.

I have not "fixed" this, because both available fixes are your call and the repo has no precedent either way:

- **(a) leave it.** #3655 merged with `Codacy = action_required` too, so the gate is evidently not blocking. Zero code churn.
- **(b) collapse the harness.** Fold the sample into the helper — `network_stats(networks, old_stats=None)` builds *and* calls, returning the dict — so the 9 call sites become one. That takes it from 13 findings to 5 and arguably reads better, but it departs from the `build_fetcher` shape you specified in the issue, so I did not do it unilaterally.
- **(c) a module-level `# pylint: disable=protected-access`.** Cheapest, but it would be the first pylint pragma in the codebase, which cuts against "match existing style".

Say which you prefer and I'll push it. Meanwhile I pushed the class docstrings the sibling `test_plugin_*.py` files carry (b11fd39) — that was a separate style gap, unrelated to the 13.

Also still open from the PR description: the decision point in the issue (I implemented **(a) leave the negative delta as is**, reasoning in the description), and whether you want the companion `develop-v5` PR — the file is byte-identical there, blob `c818def749329c9b2d00171c43cb2f2739fac5f4` on both branches.

nicolargo:
Thanks for the PR @ntdatt812 

Just merged into develop. I just remove the line in the NEWS.md file because the patch will be release in Glances 4.5.7, not 4.5.6.

Issue #3582, opened 2026-06-10
Containers plugin silently returns empty dict on Ubuntu 22.04 + Docker 29.x containerd image store (overlayfs snapshotter)


rchissotti:
## Summary

On **Ubuntu 22.04 (jammy)** with **Docker Engine 29.x fresh-installed** (which defaults to the containerd image store with the `overlayfs` snapshotter), the Glances `containers` plugin silently returns an **empty dict** instead of an array of containers. No exception is raised, no log line is emitted even with `--debug`, and the Python `docker` SDK works correctly on the same host.

The same Glances version on Ubuntu 24.04 (noble) with the same Docker 29.x + containerd snapshotter works correctly.

Reproduced on multiple machines, both Glances 4.3.3 and 4.5.4. Upstream issue #3426 ("Glances 4.5.0.1 not reporting docker container details") was claimed to fix a similar symptom, but does **not** fix this specific combo.

## Environment matrix (verified on a real fleet of Intel N100 mini-PCs)

| OS | Kernel | Docker Engine | containerd | Storage driver | Glances | `glances --stdout-json containers` |
|---|---|---|---|---|---|---|
| Ubuntu 22.04 (jammy) | 6.8.0-124 | 29.5.3 (fresh) | 2.2.4 | **overlayfs** (containerd snapshotter) | 4.3.3 | `{"containers": {}}` (silent fail) |
| Ubuntu 22.04 (jammy) | 6.8.0-124 | 29.5.3 (fresh) | 2.2.4 | **overlayfs** (containerd snapshotter) | 4.5.4 | `{"containers": {}}` (silent fail) |
| Ubuntu 22.04 (jammy) | 6.8.0-111 | 29.5.3 (fresh) | 2.2.4 | **overlayfs** (containerd snapshotter) | 4.5.4 | array of 6 containers ✓ |
| Ubuntu 22.04 (jammy) | 6.8.0-111 | 29.5.3 (in-place upgrade) | 2.2.4 | `overlay2` (legacy graph driver) | 4.5.4 | array of 6 containers ✓ |
| Ubuntu 24.04 (noble) | 6.17.0-35 | 29.5.3 (fresh) | 2.2.4 | **overlayfs** (containerd snapshotter) | 4.3.3 | array of 7 containers ✓ |

[...]

PR #3671, opened 2026-08-21
fix(plugins): init containers, vms and smart stats as a list

justadityaraj:
#### Description

`/api/4/containers` returns `{}` instead of `[]` when no container engine library is importable, which is what #3582 reports.

`ContainersPlugin` exposes a list of stats but never passed `stats_init_value` to `super().__init__()`, so it inherited the base default of `{}` (`glances/plugins/plugin/model.py:59`). `get_init_value()` is what `update()` returns when it has nothing to report:

```python
if not self.watchers:
    return self.get_init_value()
if self.input_method != 'local':
    return self.get_init_value()
```

...and it is also what the base `reset()` installs. So on a machine where `import docker` fails, the plugin answers with an empty dict where the API contract says list.

**Three plugins have this, not one.** `vms` has the identical omission. `smart` is the interesting one: it declares `stats_init_value=[]` as its *own parameter default* and then drops it on the way to the parent:

```python
def __init__(self, args=None, config=None, stats_init_value=[]):
    ...
    super().__init__(args=args, config=config)   # never forwarded
```

That reads as correct and greps as correct. Only running it shows the plugin still gets `{}`. Every other list plugin — alert, amps, diskio, folders, fs, gpu, irq, mpp, network, npu, percpu, ports, processlist, sensors, wifi — already passes `stats_init_value=[]`.

**This has been seen before and read as flakiness.** `tests/test_restful.py:149` carries a workaround for exactly this symptom:

```python
if isinstance(req.json(), dict) and not req.json():
    # Specific case for plugins that can be empty on VM (like sensors, containers...)
    # They return an empty dict instead of an empty list
    continue
```

added in 4c92e1b, *"test: Fix flaky tests (allow empty dicts for list plugins)"*. After this change that `continue` can no longer fire for any plugin in the list. I have deliberately left it alone to keep the diff on the bug — happy to remove it in a follow-up if you'd like the assertion tightened.

**On the "silently" in the issue title** — worth noting for the reporter, but not something I propose to change here. `glances/logger.py:60` pins the console handler to `CRITICAL` and nothing re-raises it under `--debug`, so the `logger.warning("Error loading Docker deps Lib. Docker plugin is disabled (...)")` at `containers/engines/docker.py:27` only ever reaches the rotating log file. That line in `glances.log` is what names their actual root cause. (Their check `from glances.plugins.containers.engines.docker import DockerExtension` proves nothing either way — that import succeeds even when the plugin is disabled, because the `try/except` only sets a flag.)

**Verification**, on macOS / Python 3.12 against `develop` @ 3bda428:

```
before:  glances --stdout-json containers  ->  {"containers": {}}
after:   glances --stdout-json containers  ->  {"containers": []}
```

and the reporter's exact code path, `watchers = {}`, now returns `[]`.

`tests/test_plugin_init_value.py` is added as a regression guard across all 19 list plugins rather than just the three, so a future plugin that forgets the argument fails immediately. Against the unpatched tree it fails 6 / passes 32 — exactly the three plugins × two assertions; with the fix, 38 pass.

Full suite is unchanged either side: **715 passed → 753 passed** (+38 new), same 8 pre-existing unrelated failures both runs (2 × `test_actions_sanitize`, 6 × `test_xmlrpc` host-header). `ruff check` and `ruff format --check` clean.

#### Resume

* Bug fix: yes
* New feature: no
* Fixed tickets: 3582

### DIFF
--- a/glances/plugins/containers/__init__.py
+++ b/glances/plugins/containers/__init__.py
@@ -147,7 +147,11 @@ class ContainersPlugin(GlancesPluginModel):
     def __init__(self, args=None, config=None):
         """Init the plugin."""
         super().__init__(
-            args=args, config=config, items_history_list=items_history_list, fields_description=fields_description
+            args=args,
+            config=config,
+            items_history_list=items_history_list,
+            stats_init_value=[],
+            fields_description=fields_description,
         )
 
         # The plugin can be disabled using: args.disable_docker
--- a/glances/plugins/smart/__init__.py
+++ b/glances/plugins/smart/__init__.py
@@ -187,13 +187,13 @@ def get_smart_data(hide_attributes):
 class SmartPlugin(GlancesPluginModel):
     """Glances' HDD SMART plugin."""
 
-    def __init__(self, args=None, config=None, stats_init_value=[]):
+    def __init__(self, args=None, config=None):
         """Init the plugin."""
         if not is_admin() and args:
             disable(args, "smart")
             logger.debug("Current user is not admin, HDD SMART plugin disabled.")
 
-        super().__init__(args=args, config=config)
+        super().__init__(args=args, config=config, stats_init_value=[])
 
         self.display_curse = True
         self.hide_attributes = self._parse_hide_attributes(config)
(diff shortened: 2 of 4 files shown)


nicolargo:
Thanks for the PR @justadityaraj 

Merged into develop

justadityaraj:
> Thanks for the PR @justadityaraj
> 
> Merged into develop

Thanks @nicolargo !

PR #3676, opened 2026-08-24
fix(folders): honour folder_N_refresh instead of walking every cycle

ntdatt812:
## The bug

`FolderList.update()` guards the per-folder timer like this:

```python
if not self.first_grab and i in self.timer_folders and not self.timer_folders[i].finished():
    continue
```

`i` is a list index; `self.timer_folders` is a list of `Timer` objects. `i in self.timer_folders` is a **membership** test, not a bounds check — it asks whether the integer `i` equals one of the timers. `Timer` doesn't implement `__eq__`, so it is always `False`:

```python
>>> from glances.timer import Timer
>>> 0 in [Timer(600), Timer(600)]
False
```

So the skip branch is unreachable, and the `self.timer_folders[i].reset()` below it never runs either.

## What it costs

`folder_N_refresh` has no effect. Every monitored folder is re-walked on **every** refresh cycle (2s by default) instead of on its own schedule — and `folder_size()` is a full recursive walk, which is the whole reason the setting exists.

Measured with `folder_1_refresh=600` and `folder_size` counted:

| | walks after 4 update cycles |
|---|---|
| before | 4 |
| after | 1 |

## The fix

Use the bound the code meant: `i < len(self.timer_folders)`.

The guard is worth keeping rather than dropping — `__folder_list` is a **class** attribute while `timer_folders` is per-instance, so the folder list can outgrow the timer list and the index has to be checked.

## Tests

`tests/test_folder_list_refresh.py`, 4 cases: refresh delay honoured, folder walked again once the timer is over, timer restarted after a walk, and a folder with no timer still updated.

Mutation-checked — putting `i in self.timer_folders` back turns 2 of the 4 red:

```
FAILED test_folder_is_not_walked_again_before_its_refresh_delay
FAILED test_the_timer_is_restarted_after_a_walk
2 failed, 2 passed
```

### DIFF
--- a/glances/folder_list.py
+++ b/glances/folder_list.py
@@ -119,7 +119,7 @@ class FolderList:
         # Iter upon the folder list
         for i in range(len(self.get())):
             # Update folder size
-            if not self.first_grab and i in self.timer_folders and not self.timer_folders[i].finished():
+            if not self.first_grab and i < len(self.timer_folders) and not self.timer_folders[i].finished():
                 continue
             # Set the key (see issue #2327)
             self.__folder_list[i]['key'] = key
@@ -132,7 +132,7 @@ class FolderList:
                     )
                 )
             # Reset the timer
-            if i in self.timer_folders:
+            if i < len(self.timer_folders):
                 self.timer_folders[i].reset()
 
         # It is no more the first time...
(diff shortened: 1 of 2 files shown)


nicolargo:
Thanks for the PR @ntdatt812 

This one is tricky ! 

Merged into develop.

PR #3678, opened 2026-08-25
fix(containers): stop repeating a title fragment when several engines run

ntdatt812:
## One append in the wrong block

`build_title` assembles the containers header by appending fragments to a single list. The final append sits **outside** the branch that produces its message:

```python
if len(self.stats) > 1:
    msg = f' {len(self.stats)}'
    ret.append(self.curse_add_line(msg))
    msg = f' sorted by {sort_for_human[self.sort_key]}'
    ret.append(self.curse_add_line(msg))
if not self.views['show_engine_name']:
    msg = f' (served by {self.stats[0].get("engine", "")})'
ret.append(self.curse_add_line(msg))      # ← outside the `if`
```

With one engine the branch runs and the append is right. With several engines the branch is skipped, `msg` still holds the fragment appended just above it, and appending it again repeats it in the header.

`show_engine_name` is set when the containers come from more than one engine:

```python
if len({ct["engine"] for ct in self.stats}) > 1:
    show_engine_name = True
```

so this is exactly the Docker + Podman case.

## Measured

Driving the real `build_title`:

| containers | engines | title |
|---|---|---|
| 2 | 2 | `CONTAINERS 2 sorted by CPU consumption `**`sorted by CPU consumption`** |
| 1 | 2 | `CONTAINERS`**`CONTAINERS`** |
| 2 | 1 | `CONTAINERS 2 sorted by CPU consumption (served by docker)` ✓ |
| 1 | 1 | `CONTAINERS (served by docker)` ✓ |

After the fix the first two read `CONTAINERS 2 sorted by CPU consumption` and `CONTAINERS`, and the two correct rows are unchanged.

## The fix

Move the append inside its branch. Nothing is lost in the multi-engine case: `maybe_add_engine_name_or_pod_line` already adds a per-row **Engine** column there, which is why the header deliberately omits "(served by …)".

## Tests

Six cases in `tests/test_plugin_containers.py`, driving `build_title` with the curses helpers stubbed:

- several engines → the sort fragment appears once
- several engines with a single container → the title appears once
- one engine → `(served by docker)` still shown, with and without the count
- the sort key is named in the title (`memory consumption`, `container name`)
- a sweep asserting no fragment appears twice in any of the four combinations

Mutation-checked — putting the append back outside the branch turns 3 of the 6 red:

```
FAILED test_several_engines_do_not_repeat_the_sort_fragment
FAILED test_several_engines_with_one_container_do_not_repeat_the_title
FAILED test_no_fragment_appears_twice
3 failed, 12 passed
```

### DIFF
--- a/glances/plugins/containers/__init__.py
+++ b/glances/plugins/containers/__init__.py
@@ -344,8 +344,12 @@ class ContainersPlugin(GlancesPluginModel):
             msg = f' sorted by {sort_for_human[self.sort_key]}'
             ret.append(self.curse_add_line(msg))
         if not self.views['show_engine_name']:
+            # With several engines the name is a per-row column instead, so
+            # there is nothing to add here. The append has to stay inside the
+            # branch: `msg` still holds the previous fragment, and appending it
+            # again repeats it in the title.
             msg = f' (served by {self.stats[0].get("engine", "")})'
-        ret.append(self.curse_add_line(msg))
+            ret.append(self.curse_add_line(msg))
         ret.append(self.curse_new_line())
         return ret
 
--- a/tests/test_plugin_containers.py
+++ b/tests/test_plugin_containers.py
@@ -11,6 +11,7 @@
 
 from types import SimpleNamespace
 
+from glances.plugins.containers import ContainersPlugin
 from glances.plugins.containers.engines.docker import DockerStatsFetcher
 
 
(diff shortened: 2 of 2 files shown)


nicolargo:
Thanks for the PR @ntdatt812 

Just merged into develop.

Issue #3419, opened 2026-02-02
InfluxDB2 Exports for AMPs can mismatch types for result field


Drulikar:
**Check the bug**
Could not find a similar *open* issue. `field type conflict` does have 4 matches for closed, but none related to AMPs.

**Describe the bug**
AMPs funnel most data into just a `result` field, but an issue with this is that many existing AMPs will put a string there, but https://github.com/nicolargo/glances/blob/develop/glances/exports/export.py#L194-L202 will always try to type the data as a float if possible regardless of what influx has assigned as the type for that column. Furthermore influx is too dumb to just automatically cast a float into a string I guess. As much as I would prefer a float data type when possible, amps need to either automatically use a different field if it wants to store a different type (e.g. `result` is always a string and a new `result_float` or something is always a float). Alternatively the amp `result` field should just always be forced to be a string type even if it can cast fine into a float.

**To Reproduce**
1. Have a docker compose config to connect to influx db and have a couple amps configured that mix the result type e.g.
```
[amps]
disable=False

[influxdb2]
host=localhost
port=8086
protocol=http
org=cm13
bucket=glances
token=YOURTOKENGOESHERE

[amp_conntrack]
enable=true
refresh=30
one_line=false
# I make a string result
command=sysctl net.netfilter.nf_conntrack_count && sysctl net.netfilter.nf_conntrack_max

[amp_package0_throttle]
enable=True
refresh=30
one_line=false
# I make a float result
command=cat /rootfs/sys/devices/system/cpu/cpu0/thermal_throttle/package_throttle_count
```
2. Start Glances on linux with the following options `-q -d -C /glances/conf/glances.conf --export influxdb2` and `- "/:/rootfs:ro"` in volumes (or something similar)
3. Wait at least 30seconds
4. Open log in `/tmp/glances-root.log` (for whatever reason logs are not sent to stdout like they should be from https://github.com/nicolargo/glances/commit/34da61647b2d091757091530b207bb1fd659d563 you only get some python errors about `SyntaxWarning: invalid escape sequence`) e.g.:

Docker logs:
```
sudo docker compose logs
glances-1  | /venv/lib/python3.12/site-packages/pymdstat/pymdstat.py:107: SyntaxWarning: invalid escape sequence '\W'
glances-1  |   return [split('\W+', i)[1] for i in line.split(':')[1].split(' ') if i.startswith('[')]
[...]

PR #3679, opened 2026-08-25
feat(export): publish a numeric AMP result as result_float too (#3423)

### LINKED ISSUE #3423: Add a new result_float field in AMPs plugins for InfluxDB export
See enhancement in https://github.com/nicolargo/glances/issues/3419

ntdatt812:
Closes #3423.

## The comment describes a field that does not exist

`normalize_for_influxdb` says:

```python
# Some fields should be converted to string in order to avoid type mismatch in InfluxDB
# Example: the 'result' field of the AMP plugin can be a string or a number depending on the AMP implementation.
# In this case, we convert it to a string and create another field with the same name but with a suffix (_float)
# to keep the original value.
# See #3419 for more details.
FIELD_TO_STRING = ["result"]
```

Only the first half is implemented:

```python
if k in FIELD_TO_STRING:
    fields[k] = str(fields[k])
```

`grep -rn "_float" glances/` finds the comment and nothing else — the companion field was never created.

## Why it matters

Forcing `result` to a string is what stopped InfluxDB rejecting the second AMP that wrote a different type into the same column (#3419). But it also left every **numeric** AMP with no numeric field at all: a `package_throttle_count` that used to be graphable is now a string, and there is no way to chart it.

Measured against the real `normalize_for_influxdb`:

| AMP result | before | after |
|---|---|---|
| `'42'` | `result: str '42.0'` | `result: str '42.0'`, **`result_float: float 42.0`** |
| `3.5` | `result: str '3.5'` | `result: str '3.5'`, **`result_float: float 3.5`** |
| `'nf_conntrack_count = 123'` | `result: str …` | `result: str …` *(no companion)* |

A textual result publishes no companion — there is no number to publish — so the two AMPs from the original report still agree on the type of every column they share. `result` itself is untouched, so existing dashboards keep working.

## Two details

- The companion is collected during the walk and merged afterwards. A dict cannot grow while it is being iterated, and writing it inline would raise `RuntimeError: dictionary changed size during iteration` the first time a numeric AMP reported.
- `NaN` and `inf` are excluded. InfluxDB rejects them, and one would fail the **whole** write — a worse outcome than the missing field this change is fixing.

## Tests

`tests/test_export_influxdb_normalize.py` — 11 cases:

- a numeric result is published as a float, and `result` stays a string
- a textual result publishes no companion
- the string AMP and the numeric AMP from #3419 agree on `result`'s type, and the numeric one still gets its float
- `nan`, `inf`, `-inf` publish no companion
- `None` publishes neither field; `0` still publishes `0.0` (not swallowed as falsy)
- ordinary numeric fields keep their float type and get **no** companion — only `FIELD_TO_STRING` names do
- a measurement without a `result` is byte-for-byte unchanged

Mutation-checked, twice:

```
# drop the companion field
× a_numeric_result_is_also_published_as_a_float
× the_two_amps_from_the_report_can_coexist
× a_zero_result_is_still_published                     3 failed, 8 passed

# drop the isfinite guard
× non_finite_results_are_not_published_as_floats[nan]
× ...[inf]
× ...[-inf]                                            3 failed, 8 passed
```

### DIFF
--- a/glances/exports/export.py
+++ b/glances/exports/export.py
@@ -12,6 +12,7 @@ I am your father...
 """
 
 import re
+from math import isfinite
 
 from glances.globals import NoOptionError, NoSectionError, json_dumps
 from glances.logger import logger
@@ -172,9 +173,10 @@ class GlancesExport:
         FIELD_TO_TAG = ["name", "cmdline", "type"]
         # Some fields should be converted to string in order to avoid type mismatch in InfluxDB
         # Example: the 'result' field of the AMP plugin can be a string or a number depending on the AMP implementation.
-        # In this case, we convert it to a string and create another field with the same name but with a suffix (_float)
-        # to keep the original value.
-        # See #3419 for more details.
+        # In this case, we convert it to a string and, when the value is numeric,
+        # publish it a second time under a '<name>_float' field so it stays
+        # queryable as a number.
+        # See #3419 and #3423 for more details.
         FIELD_TO_STRING = ["result"]
 
         # Build initial dict by crossing columns and point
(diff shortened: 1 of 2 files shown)


nicolargo:
Thanks for the PR @ntdatt812  but as it break the API i will implement it directly in the Glances v5 branches.

For the moment , i close this one.

PR #3680, opened 2026-08-25
fix(processlist): show Windows priority classes as labels in the NI column

### LINKED ISSUE #3672: [Windows] In TUI and WebUI, Nice level is not displayed correctly

ntdatt812:
Closes #3672.

## The measurement

`psutil` on Windows returns the priority class, and the `NI` column is `'{:>3} '`. Two of the six classes are five digits, so they render six characters into a four-character cell and shift every column to their right. Measured on this machine (Windows 11, 519 processes):

```
REALTIME_PRIORITY_CLASS       -> '256 '
HIGH_PRIORITY_CLASS           -> '128 '
ABOVE_NORMAL_PRIORITY_CLASS   -> '32768 '   <- 6 chars in a 4-char column
NORMAL_PRIORITY_CLASS         -> ' 32 '
BELOW_NORMAL_PRIORITY_CLASS   -> '16384 '   <- 6 chars in a 4-char column
IDLE_PRIORITY_CLASS           -> ' 64 '
```

They are not hypothetical. A live count on this box:

```
NORMAL_PRIORITY_CLASS         324 processes
None                          146
IDLE_PRIORITY_CLASS            33
ABOVE_NORMAL_PRIORITY_CLASS     8   <- overflows
BELOW_NORMAL_PRIORITY_CLASS     5   <- overflows
HIGH_PRIORITY_CLASS             3
```

After, every class is four characters:

```
 RT   Hi   AN   No   BN   Lo
```

## What changed

**TUI** — `nice_to_str()` and `WINDOWS_PRIORITY_LABELS` in `glances/plugins/processlist/__init__.py`, applied at format time only.

**WebUI** — the same table and the same labels in `plugin-processlist.vue`, as `niceLabel()`. The component already carried Windows-specific nice knowledge (`isWindows && process.nice != 32` for the `isNice` highlight), so the table sits beside it rather than being routed through the API; the comment on each copy points at the other.

Two decisions worth stating:

- **An unknown class is left as its number, not blanked.** If psutil starts reporting a class this table does not know, showing `12345` is honest; showing nothing hides it.
- **The alert still reads the raw value.** `get_nice_alert` compares against `nice_careful` / `nice_warning` / `nice_critical` from `glances.conf`, which are numbers. Handing it `'AN'` would silently disable every threshold on Windows. There is a test for exactly this.

I used the labels from the issue. Note `64` is `IDLE_PRIORITY_CLASS` in psutil while the issue calls it *Low*; I kept your `Lo` since that is the Windows-facing name, but say the word and I will change it.

## Tests

`tests/test_plugin_processlist_nice.py`, 24 tests, both platform branches via `patch.object(processlist_module, 'WINDOWS', ...)` so they run everywhere:

```
24 passed in 0.13s
```

Mutation-checked, not just written:

| mutation | result |
|---|---|
| `nice_to_str` returns `nice` unconditionally (verified the `WINDOWS_PRIORITY_LABELS.get` call count went 1 → 0) | **13 failed**, 11 passed |
| alert fed `nice_to_str(nice)` instead of the raw value | **1 failed** — `test_alert_reads_the_raw_value_not_the_label` |

Both reverted; back to 24 passed.

Wider suite, unchanged: `test_core.py`, `test_plugin_processcount.py`, `test_plugin_model.py`, `test_processes_cache.py` plus the new file → **113 passed, 6 skipped**.

## The bundle

`glances/outputs/static/public/glances.js` is rebuilt, because it is committed and a `.vue` change that skips it ships only half the fix.

To show I did not overwrite anyone's build, I rebuilt **from unmodified source first**: `git status glances/outputs/static/public/` came back empty, so the build is reproducible on this machine and the only delta below is mine. Then rebuilt with the change — `glances.js` 686564 → 686737 bytes, `browser.js` untouched. The minified output carries it:

```js
const SA={256:"RT",128:"Hi",32768:"AN",32:"No",16384:"BN",64:"Lo"};
function EA(t,e){return e&&null!=t?SA[t]??t:t}
```

## Lint

- `ruff check` clean, `ruff format --check` clean on both Python files.
- `eslint ./js/components/plugin-processlist.vue`: **5 errors / 75 warnings before, 5 errors / 75 warnings after** — measured by stashing the change and re-running. All pre-existing; zero net new.

## One thing I noticed but did not touch

146 processes on this box report `nice = None`. The TUI turns that into `?`; the WebUI passes it to `$filters.exclamation`, which only maps `undefined` and `""` to `?`, so `null` renders as an empty cell. Different symptom, different fix — happy to send it separately if you want it.


ntdatt812:
Superseded by #3673, which is merged. That one already carries the same labels on both WebUI render paths and the TUI side, so there is nothing left here to apply — I checked `plugin-processlist.vue` and `processlist/__init__.py` on the current default branch before closing. Closing to keep the queue honest; #3672 stays fixed by #3673.

PR #3681, opened 2026-08-25
fix(webui): show ? for a process whose nice value the OS will not report

ntdatt812:
Found while measuring #3672; reported there at the time and split out rather than folded in, since it is a different defect in a different file.

## The measurement

`exclamation` maps `undefined` and `""` to `?`, and nothing else. The API sends JSON, so a process whose nice value the OS will not report reaches the WebUI as **`null`** — and Vue renders `null` as an empty string.

Measured by pulling the real function body out of `js/filters.js` and feeding it Vue's own `toDisplayString`:

| `process.nice` | before | after |
|---|---|---|
| `undefined` | `"?"` | `"?"` |
| **`null`** | **`""`** | **`"?"`** |
| `""` | `"?"` | `"?"` |
| `0` | `"0"` | `"0"` |
| `-20` | `"-20"` | `"-20"` |
| `32` | `"32"` | `"32"` |

Not a rare shape. On this box, through glances' own collector:

```
processes: 513
  nice is None : 144
  no nice key  : 0
  JSON the WebUI receives: {"nice": null}
  sample names: ['MemCompression', 'MsMpEng.exe', 'sqlservr.exe', '']
```

144 of 513 rows — 28% — render a blank NI cell in the WebUI while the TUI shows `?` for the same process. `_get_process_curses_nice` substitutes `'?'` for `None` before formatting; the WebUI had no equivalent.

## The change

One condition, plus a comment saying why it is written the way it is.

**Explicit comparisons, not a falsy test.** `0` is a real nice value on every POSIX system — the default one — and `if (!input)` would turn every normal-priority process on Linux into a `?`. That row is in the table above precisely because it is the trap here.

## Scope

`exclamation` has exactly two call sites, `plugin-processlist.vue:186` and `:365`, both the NI column. Nothing else changes shape.

No JS test framework exists in `glances/outputs/static`, so there is no unit test to add here; the measurement above is against the committed source rather than a copy of it.

## The bundle

`public/glances.js` and `public/browser.js` are rebuilt — `filters.js` is shared, so both carry it.

Rebuilt **from unmodified source first** to show the build is reproducible on this machine and the delta below is only mine: `git status glances/outputs/static/public/` came back empty. Then rebuilt with the change — `glances.js` 686564 → 686561, `browser.js` 573037 → 573034 bytes.

The minified function, before and after:

```js
// HEAD
(t){return void 0===t||""===t?"?":t}
// this branch
(t){return null==t||""===t?"?":t}
```

`null==t` is the minifier collapsing the two explicit comparisons into one — it covers `null` and `undefined` both, which is exactly the intent.

`eslint ./js/filters.js`: clean.

## Note on #3680

That PR also rebuilds `public/glances.js` (for the Windows priority labels in the same column), so whichever of the two lands second will need its bundle rebuilt on top. Say the word and I will rebase and rebuild — or fold this into #3680 if you would rather review the NI column as one change.


ntdatt812:
Rebased onto `develop` — the conflict was only in the two committed bundles, which had moved on since I opened this. The source change is unchanged.

Both bundles rebuilt from the current `develop`, and the whole diff is one operator:

```
glances.js : "void 0=" -> "null"
browser.js : "void 0=" -> "null"
```

Everything before and after that region is byte-identical, so the rebuild carried nothing else. The minifier collapsed `void 0===t||null===t` into `null==t` — which is why both files got *smaller* by 3 bytes despite the source gaining a condition. Loose `==` against `null` is exactly right here: it is the one comparison that matches `null` and `undefined` and nothing else, which is the intent.

Recapping the fix itself, since it has been sitting a while: the API sends JSON, so a process whose nice value the OS will not report arrives as `null`, and Vue renders `null` as an empty string — a blank cell where the TUI shows `?`. The comparisons stay explicit rather than a falsy test, because `0` is a real nice value on every POSIX system and has to render as `0`.

nicolargo:
Thanks @ntdatt812 

Merged into develop.

PR #3682, opened 2026-08-25
fix(webui): colour the per-CPU iowait cell from iowait, not system

ntdatt812:
Three things in `plugin-percpu.vue`, one of which shows the operator a wrong colour.

## The iowait column takes its colour from `system`

```js
getIOWaitAlert(cpu) {
  return GlancesHelper.getAlert("percpu", "percpu_iowait_", cpu.system);
}
```

It looks up the `percpu_iowait_*` thresholds and then measures `cpu.system` against them. Its two neighbours each read their own field — `getUserAlert` uses `cpu.user`, `getSystemAlert` uses `cpu.system` — so this is the only one out of step.

What an operator sees, with the shipped defaults (careful 50 / warning 70 / critical 90):

| `system` | `iowait` | iowait cell today | should be |
|---|---|---|---|
| 95% | 2% | **CRITICAL** (red) | OK |
| 5% | 95% | OK | **CRITICAL** (red) |

So the column is red exactly when the disks are fine, and plain exactly when iowait is the thing to look at — the wrong answer in both directions rather than merely a missing one. It also silently ignores `percpu_iowait_*` being tuned separately from `percpu_system_*`, which the config lets you do.

It is in the shipped bundle, not only the source:

```
$ git show HEAD:glances/outputs/static/public/glances.js | grep -o 'percpu_iowait_.\{0,25\}'
percpu_iowait_",t.system)}}
```

and after the rebuild in this PR:

```
percpu_iowait_",t.iowait)}}
```

**The TUI was never affected.** `display_cpu_stats_in_columns` does `self.get_alert(cpu[stat], header=stat)` — the value and the threshold come from the same `stat`. This is a WebUI-only divergence.

## `steel` → `steal`

The header cell for CPU steal time read `steel`. `grep -c steel` on the rebuilt bundle is now 0.

## Headers for columns this platform never fills

`idle`, `iowait` and `steal` printed a header unconditionally, while their value cells were gated on the field being defined. psutil reports `('user', 'system', 'idle', 'interrupt', 'dpc')` on Windows and no `iowait`/`steal` on macOS either, so those platforms drew header cells with nothing underneath. Verified here on Windows 11:

```
>>> psutil.cpu_times_percent(percpu=True)[0]._fields
('user', 'system', 'idle', 'interrupt', 'dpc')
```

Header and value now ask the same question through one `hasStat(name)` helper, so they cannot drift apart. It asks the whole table rather than a single row — the previous per-row `v-show` would have given rows different widths had one CPU ever reported a field another did not.

## Verification

- WebUI rebuilt with `npm run build`; the diff is the `.vue` source plus the regenerated `public/glances.js`. The bundle keeps LF endings and the same 48-line shape as before, so the only change in it is the compiled component.
- `npx eslint ./js/components/plugin-percpu.vue` reports the same two findings before and after this change (`vue/require-default-prop`, `@typescript-eslint/no-unused-vars`) — both pre-existing on `develop`, neither introduced here.
- No Python changed, so the plugin's own tests are untouched.

There is no JS unit-test runner in `glances/outputs/static` (the `package.json` scripts are build/start/watch/lint/clean), which is why the evidence above is the before/after bundle rather than a new test. Happy to add one if you would like a runner introduced.


ntdatt812:
Closing in favour of #3698, which fixes the same bug — `getIOWaitAlert` passing `cpu.system` to the iowait threshold lookup.

I opened this one on the 25th and then, not spotting it, opened #3698 for the same thing. My error. Keeping #3698 rather than this one because it is the better of the two:

- this branch has gone `CONFLICTING` on the rebuilt `public/glances.js`, #3698 applies cleanly to `develop`
- #3698 also corrects the column header, which reads `steel` instead of `steal`
- #3698 carries the bundle proof: `cmp -l` shows exactly 7 bytes changed, so the rebuild demonstrably carried nothing else

Sorry for the noise.

PR #3683, opened 2026-08-25
fix(connections): alert on the tracked percentage, not on 0

ntdatt812:
`connections/__init__.py` builds the conntrack decoration like this:

```python
self.views['nf_conntrack_percent']['decoration'] = self.get_alert(header='nf_conntrack_percent')
```

No value is passed, so `get_alert` uses its own default — `current=0` — and computes `value = 0 * 100 / 100`. Zero is below every threshold, so the result is `OK` on every refresh regardless of how full the table is.

The thresholds are real and documented. `conf/glances.conf` ships them:

```ini
# nf_conntrack thresholds in %
nf_conntrack_percent_careful=70
nf_conntrack_percent_warning=80
nf_conntrack_percent_critical=90
```

and `docs/aoa/connections.rst` documents the same three. None of them could ever fire.

That matters more than a missing colour: once `nf_conntrack_count` reaches `nf_conntrack_max` the kernel starts dropping new connections, which is precisely the condition a `critical=90` threshold is there to warn about before it happens. `nf_conntrack_percent` is already computed correctly in `update_for_nf_conntrack_method`, so the number on screen was right the whole time — only its colour was decided from 0.

Both interfaces read this one decoration — the WebUI through `getDecoration('nf_conntrack_percent')` in `plugin-connections.vue` — so both were silent, and both are fixed by the one change.

## Tests

Added to `tests/test_connections_states.py`, reusing that file's plugin fixture. `update_views` runs over injected stats and the decoration is read back:

- 95% → `CRITICAL`, 85% → `WARNING`, 75% → `CAREFUL`
- 10% → `OK`

The limits are set on the fixture to the values `conf/glances.conf` ships, since the fixture builds the plugin with `config=None`.

**Mutation-checked** — restoring `get_alert(header='nf_conntrack_percent')` fails exactly one test, on the first assertion:

```
- CRITICAL
+ OK
1 failed, 5 passed
```

The `10% → OK` case deliberately does not catch that mutation: it pins the low end, which reads `OK` either way. It is there so a future change cannot fix the high end by colouring everything.

```
python -m pytest tests/test_connections_states.py -q
6 passed
```

No WebUI rebuild is needed — the change is Python only, and the bundle already reads the decoration the plugin publishes.

### DIFF
--- a/glances/plugins/connections/__init__.py
+++ b/glances/plugins/connections/__init__.py
@@ -182,7 +182,9 @@ class ConnectionsPlugin(GlancesPluginModel):
         try:
             # Alert and log
             if self.stats['nf_conntrack_enabled']:
-                self.views['nf_conntrack_percent']['decoration'] = self.get_alert(header='nf_conntrack_percent')
+                self.views['nf_conntrack_percent']['decoration'] = self.get_alert(
+                    self.stats['nf_conntrack_percent'], header='nf_conntrack_percent'
+                )
         except KeyError:
             # try/except mandatory for Windows compatibility (no conntrack stats)
             pass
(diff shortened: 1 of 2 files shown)


nicolargo:
Thanks for the PR @ntdatt812 

Merged into develop

PR #3684, opened 2026-08-25
fix(diskio): alert on the bitrate, and decorate the field the WebUI reads

ntdatt812:
Two things, one line apart, and the network plugin next door already does both correctly.

## The alert measured a lifetime counter against a bitrate threshold

```python
alert_rx = self.get_alert(i['read_bytes'], header='rx', action_key=disk_real_name)
```

`read_bytes` carries `rate: True` in `fields_description`, which is what produces `read_bytes_rate_per_sec` — `read_bytes` itself is psutil's counter since boot. It only ever grows, so on any machine with real uptime a configured `..._rx_critical` was crossed within seconds of the first read and stayed crossed for the life of the process, regardless of what the disk was doing. An operator who set thresholds got a permanently red column; one who set none saw nothing, which is why this could sit unnoticed.

The network plugin is the shape this should have had:

```python
bps_rx = int(i['bytes_recv_rate_per_sec'] * 8)
alert_rx = self.get_alert(bps_rx, header='rx', action_key=if_real_name)
```

## The decoration went to a key one interface reads and the other does not

`update_views` set it only on `read_bytes` / `write_bytes`:

- the **TUI** prints `read_bytes_rate_per_sec` (`auto_unit(i.get('read_bytes_rate_per_sec'))`) but reads `read_bytes`' decoration — a colour computed from a different quantity than the number beside it;
- the **WebUI** asks for `getDecoration(disk.name, 'read_bytes_rate_per_sec')`, which nothing ever set, so it rendered no colour at all.

Network sets both keys for exactly this reason:

```python
self.views[...]['bytes_recv']['decoration'] = alert_rx
self.views[...]['bytes_recv_rate_per_sec']['decoration'] = alert_rx
```

Diskio now does the same, guarded: the `_rate_per_sec` view only exists once there is a timespan to divide by, and setting it unconditionally raises `KeyError` on the first sample — the repo's existing `TestDiskioPluginViews` cases catch that, which is how I found it.

## Tests

`tests/test_plugin_diskio.py`, new `TestDiskioBitrateAlert`, four cases over injected stats carrying a large lifetime counter and a separate rate:

- quiet disk (rate 10) with 5 GB of history → `OK` — this is the case the old code got wrong
- busy disk (rate 95) → `CRITICAL`
- `read_bytes_rate_per_sec` decorated `CRITICAL` while `write_bytes_rate_per_sec` is `OK` — the WebUI's key, and asymmetric so a blanket assignment cannot pass it
- first sample with no rate field at all → no `KeyError`

Thresholds in the fixture are named `diskio_sda_rx_*`: `get_stat_name()` composes plugin + `action_key` + header, so a per-disk threshold carries the disk name, exactly as the network plugin's docstring documents with `network_wlan0_rx_careful`.

**Mutation-checked**, one per defect:

| mutation | result |
|---|---|
| alert on `i['read_bytes']` again | 2 fail — the quiet-disk case and the WebUI-key case |
| stop setting `*_rate_per_sec` | 1 fail — only the WebUI-key case |

```
python -m pytest tests/test_plugin_diskio.py -q
41 passed
```

(37 before this PR, plus the 4 added.) Python only, so no WebUI rebuild — the bundle already reads whatever the plugin publishes.

### DIFF
--- a/glances/plugins/diskio/__init__.py
+++ b/glances/plugins/diskio/__init__.py
@@ -189,11 +189,22 @@ class DiskioPlugin(GlancesPluginModel):
             # if not i.get('read_bytes_rate_per_sec') or not i.get('write_bytes_rate_per_sec'):
             #     continue
 
-            # Decorate the bitrate with the configuration file
-            alert_rx = self.get_alert(i['read_bytes'], header='rx', action_key=disk_real_name)
-            alert_tx = self.get_alert(i['write_bytes'], header='tx', action_key=disk_real_name)
-            self.views[i[self.get_key()]]['read_bytes']['decoration'] = alert_rx
-            self.views[i[self.get_key()]]['write_bytes']['decoration'] = alert_tx
+            # Decorate the bitrate with the configuration file.
+            # The thresholds describe a bitrate, so they are measured against the rate and
+            # not against read_bytes/write_bytes, which are lifetime counters: those only
+            # grow, so any configured threshold latched on and stayed there. Same shape as
+            # the network plugin, which alerts on bytes_recv_rate_per_sec.
+            alert_rx = self.get_alert(i.get('read_bytes_rate_per_sec') or 0, header='rx', action_key=disk_real_name)
+            alert_tx = self.get_alert(i.get('write_bytes_rate_per_sec') or 0, header='tx', action_key=disk_real_name)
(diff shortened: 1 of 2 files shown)


nicolargo:
Thanks for the PR @ntdatt812 

Merged into develop.

PR #3686, opened 2026-08-25
fix(WebUI): Make container IO and network columns sortable

### LINKED ISSUE #3662: Web UI: sort containers by IO and RX/TX

Glances web ui is perfect companion to quickly see what's going on with your containers. We can sort the container list by CPU, MEM and Name. But some metrics can not, even if this may provide meaningful insights to quickly see the network or IO eaters. 

IPRps, IOWps, RXps, and TXps are sortable (ASC/DESC) in container list in Web UI. 

**Describe alternatives you've considered**

none. 

**Additional context**

Thanks for the great tool.

justadityaraj:
## Problem

The container table exposes disk read/write and network receive/transmit rates, but those columns cannot be selected as sort keys. This makes it harder to identify the busiest containers from the Web UI.

Fixes https://github.com/nicolargo/glances/issues/3662

## Changes

- Make the IORps, IOWps, RXps, and TXps headers use the existing sortable-column interaction.
- Add the four underlying rate fields to the container sort allowlist and give each a readable sort label.

Only `js/components/plugin-containers.vue` is touched. The generated `public/glances.js` bundle is left untouched for the release build.

## User impact

Users can click any container IO or network-rate header to sort the table by that metric, with the active column shown using the same styling and summary used for CPU, memory, and name.

## Verification

- `npx eslint js/components/plugin-containers.vue`: 0 errors; the component's 7 pre-existing warnings remain.
- `git diff --check`: clean.

### DIFF
--- a/glances/outputs/static/js/components/plugin-containers.vue
+++ b/glances/outputs/static/js/components/plugin-containers.vue
@@ -89,10 +89,38 @@
                             MEM
                         </td>
                         <td v-show="!getDisableStats().includes('mem')" scope="col">MAX</td>
-                        <td v-show="!getDisableStats().includes('diskio')" scope="col">IORps</td>
-                        <td v-show="!getDisableStats().includes('diskio')" scope="col">IOWps</td>
-                        <td v-show="!getDisableStats().includes('networkio')" scope="col">RXps</td>
-                        <td v-show="!getDisableStats().includes('networkio')" scope="col">TXps</td>
+                        <td
+                            v-show="!getDisableStats().includes('diskio')"
+                            scope="col"
+                            :class="['sortable', sorter.column === 'io_rx' && 'sort']"
+                            @click="args.sort_processes_key = 'io_rx'"
+                        >
+                            IORps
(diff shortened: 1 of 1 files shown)


nicolargo:
Hi @justadityaraj 

some conflict appears in the glances.js file.

This file is dynamically generated so it should not be change by PR (only during release after all merges).

Can you rework this one ?

justadityaraj:
Reworked, thanks for the pointer. The branch is rebased on latest develop as a single commit that only touches js/components/plugin-containers.vue. The generated public/glances.js is no longer part of the PR, so the conflict is gone.

nicolargo:
Thanks for the quick fix @justadityaraj 

Just merged into develop !

I will add this information (never commit glances.js file in PR in the SKILL file).

justadityaraj:
> Thanks for the quick fix @justadityaraj
> 
> Just merged into develop !
> 
> I will add this information (never commit glances.js file in PR in the SKILL file).

Thanks!

PR #3687, opened 2026-08-26
fix(percpu): summarize the cores that are not displayed, not the ones that are

ntdatt812:
## The defect

`summarize_all_cpus_not_displayed` slices the list it was handed like this (`glances/plugins/percpu/__init__.py:214`):

```python
percpu_stats = [i[stat] for i in percpu_list[0 : self.max_cpu_display]]
```

That is the **same slice `msg_curse` printed one line above** (`:248`):

```python
for cpu in percpu_list[0 : self.max_cpu_display]:
    ...
# Add a new line with sum of all others CPU
return self.summarize_all_cpus_not_displayed(percpu_list, header, stats_added)
```

So the `CPU*` row averages the cores that **are** displayed, never the ones that are not — contradicting its own name and the comment at the call site.

It matters more than a duplicated number, because `manage_max_cpu_to_display` sorts descending by `total` whenever the list overflows (`:175`). The displayed slice is therefore the **busiest** cores, so `CPU*` systematically over-reports: the busier the top cores, the more misleading the summary of the quiet ones.

## Measured on `develop`

8 cores, `max_cpu_display = 4`:

```
displayed CPUs      : [90.0, 88.0, 86.0, 84.0]
NOT displayed CPUs  : [2.0, 2.0, 2.0, 2.0]
CPU* row shows      : 87.0   <- mean of the four already on screen
CPU* row should show: 2.0    <- mean of the four it claims to summarize
same slice as the displayed loop: True
```

Rendering the real row confirms it end to end: `'\nCPU*   87.0%'` for a group of cores sitting at 2%.

## The fix

```diff
-                percpu_stats = [i[stat] for i in percpu_list[0 : self.max_cpu_display]]
+                percpu_stats = [i[stat] for i in percpu_list[self.max_cpu_display :]]
```

One slice. The displayed slice and the summarized slice now partition the list — every core is either shown or summarized, exactly once, none twice.

The `len(self.stats) > self.max_cpu_display` guard already wrapping the block means the tail is never empty, so the existing `sum(...) / len(...)` needs no extra protection.

## Tests

New `tests/test_plugin_percpu.py`, 6 cases, all driving the real `summarize_all_cpus_not_displayed` and parsing the rendered row rather than re-implementing the slicing:

- the summary is the leftover mean, not the displayed mean;
- holding the leftovers fixed while changing the displayed cores does not move the summary;
- it averages rather than sums;
- a single leftover core is reported as itself;
- one percentage per header column;
- no row at all when every core fits.

**5 of the 6 fail on `develop` and pass with the patch**; the sixth guards the "everything fits" branch, which this change does not touch.

```
$ pytest tests/test_plugin_percpu.py -q
6 passed

$ pytest tests/test_plugin_cpu.py tests/test_plugin_percpu.py -q
33 passed, 1 skipped
```

### DIFF
--- a/glances/plugins/percpu/__init__.py
+++ b/glances/plugins/percpu/__init__.py
@@ -211,7 +211,7 @@ class PercpuPlugin(GlancesPluginModel):
                 return_.append(self.curse_add_line('CPU* '))
 
             for stat in header:
-                percpu_stats = [i[stat] for i in percpu_list[0 : self.max_cpu_display]]
+                percpu_stats = [i[stat] for i in percpu_list[self.max_cpu_display :]]
                 cpu_stat = sum(percpu_stats) / len(percpu_stats)
                 try:
                     msg = f'{cpu_stat:6.1f}%'
--- a/tests/test_plugin_percpu.py
+++ b/tests/test_plugin_percpu.py
@@ -1,0 +1,83 @@
+#!/usr/bin/env python
+#
+# Glances - An eye on your system
+#
+# SPDX-FileCopyrightText: 2026 Nicolas Hennion <nicolas@nicolargo.com>
+#
+# SPDX-License-Identifier: LGPL-3.0-only
+#
+
+"""Tests for the PerCPU plugin."""
+
+import re
+from unittest.mock import MagicMock
+
+import pytest
+
+from glances.plugins.percpu import PercpuPlugin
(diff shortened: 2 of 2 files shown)


nicolargo:
Thanks @ntdatt812 

Merged into develop !

PR #3690, opened 2026-08-26
fix(webui): match the RAID alert to raid_alert instead of a bare comparison

ntdatt812:
## The defect

The WebUI decides RAID degradation with one comparison (`glances/outputs/static/js/components/plugin-raid.vue:68`):

```js
degraded: diskData.used < diskData.available,
```

The authoritative implementation on the Python side applies three rules **before** reaching that comparison (`glances/plugins/raid/__init__.py:157`):

```python
if raid_type == 'raid0':
    return 'OK'
if status == 'inactive':
    return 'CRITICAL'
if used is None or available is None:
    return 'DEFAULT'
if used < available:
    return 'WARNING'
return 'OK'
```

Two of the omissions are visible to the user:

**raid0.** It has no redundancy, so `used < available` carries no meaning there — `msg_curse` excludes it from the very same "Degraded mode" banner (`:143`, `array_stats['type'] != 'raid0' and ...`). The WebUI drew the banner and the warning colour anyway.

**Unknown device count.** `used` is `None` whenever the parser cannot read it — which is exactly why `raid_alert` guards it. In JavaScript `null < 5` is **true**, so an array whose device count could not be read was reported as degraded:

```
used=null available=5   ->  degraded=true
used=undefined available=5  ->  degraded=false
```

The TUI leaves that array undecorated.

## Measured

Both implementations run over the same seven arrays:

```
case                               python    webui old   webui new   verdict
raid0, fewer used than available   ok        warning     ok          MISMATCH -> OK
raid1 genuinely degraded           warning   warning     warning     OK -> OK
raid1 healthy                      ok        ok          ok          OK -> OK
raid1 inactive                     critical  critical    critical    OK -> OK
device count unreadable            (none)    warning     (none)      MISMATCH -> OK
available unreadable               (none)    ok          (none)      MISMATCH -> OK
raid0 inactive                     ok        critical    ok          MISMATCH -> OK
```

Four disagreed before, none after. The Python column calls the real `RaidPlugin.raid_alert`.

## One asymmetry preserved, not fixed

`raid_alert` returns `OK` for an **inactive raid0**, because the raid0 branch runs before the inactive check. This change matches that, so the two views agree — but I want to flag it rather than quietly change it: whether that ordering is what you intend is a separate call, and the WebUI renders the `└─ Status inactive` line regardless of the colour. Say the word and I will swap the two branches on both sides in a follow-up.

## Verification

There is no JS test runner in this repo, so the evidence is the comparison table above plus the rebuilt bundle:

```
$ grep -c raid0 glances/outputs/static/public/glances.js
develop : 0
patched : 1
```

`glances.js` rebuilt with `npm run build` (webpack 5.107.2, compiled successfully), 686767 → 686883 bytes, and committed alongside the source as the other WebUI changes in this repo do.

ESLint on the changed file: 0 errors. The single `vue/require-default-prop` warning is pre-existing on `develop` and untouched.


nicolargo:
Thanks for the PR @ntdatt812 

Merged into develop

Issue #1373, opened 2018-12-18
TypeError on opening (Wifi plugin)


G4Zz0L1:
#### Description

At the start of Glances it errors out with this. It started today, but I see nothing 'pythonish' in the pacman update log. The debug file seems to contains nothing useful, but I'll attach it down below. I don't know where to look for other errors.

```
Traceback (most recent call last):
  File "/usr/bin/glances", line 11, in <module>
    load_entry_point('Glances==3.0.2', 'console_scripts', 'glances')()
  File "/usr/lib/python3.7/site-packages/glances/__init__.py", line 140, in main
    start(config=config, args=args)
  File "/usr/lib/python3.7/site-packages/glances/__init__.py", line 106, in start
    mode = GlancesMode(config=config, args=args)
  File "/usr/lib/python3.7/site-packages/glances/standalone.py", line 73, in __init__
    self.stats.update()
  File "/usr/lib/python3.7/site-packages/glances/stats.py", line 226, in update
    self._plugins[p].update_views()
  File "/usr/lib/python3.7/site-packages/glances/plugins/glances_wifi.py", line 155, in update_views
    self.views[i[self.get_key()]]['signal']['decoration'] = self.get_alert(i['signal'])
  File "/usr/lib/python3.7/site-packages/glances/plugins/glances_wifi.py", line 136, in get_alert
    if value <= self.get_limit('critical', stat_name=self.plugin_name):
TypeError: '<=' not supported between instances of 'NoneType' and 'float'

```

#### Versions

* Glances & psutil (glances -V):
Glances v3.0.2 with psutil v5.4.8
* Operating System (lsb_release -a):
LSB Version:	1.4
Distributor ID:	Arch
Description:	Arch Linux
Release:	rolling
Codename:	n/a

#### Logs

Glanced debug log
https://pastebin.com/sbdJSgTi

nicolargo:
Should be corrected by the latest commit (229710d6) on the DEVELOP branch.

@G4Zz0L1  Can you test it ?

Thanks

nicolargo:
Another day, another patch.

Can you test it agaib @G4Zz0L1 ?

G4Zz0L1:
It seems to work now, but it doesn't fetch wifi info

![image](https://user-images.githubusercontent.com/13220636/50384312-6b4b5800-06c3-11e9-87a4-a50f2932668e.png)

nicolargo:
[...]

PR #3691, opened 2026-08-26
fix(wifi): honour a partially configured set of signal thresholds

ntdatt812:
## The defect

`WifiPlugin.get_alert` overrides the base implementation and chains its three comparisons (`glances/plugins/wifi/__init__.py:146`):

```python
try:
    if value <= self.get_limit('critical', stat_name=self.plugin_name):
        ret = 'CRITICAL'
    elif value <= self.get_limit('warning', stat_name=self.plugin_name):
        ret = 'WARNING'
    elif value <= self.get_limit('careful', stat_name=self.plugin_name):
        ret = 'CAREFUL'
except (TypeError, KeyError):
    # Catch TypeError for issue1373
    ret = 'DEFAULT'
```

`get_limit` returns `None` for a level the config does not define (`plugin/model.py:969`). `value <= None` raises `TypeError`, the blanket handler catches it, and the whole result collapses to `DEFAULT`.

Because `critical` is tested **first**, omitting it silences the other two as well.

## Measured on `develop`

```
all three thresholds configured:
  quality_level= -50 dBm -> OK
  quality_level= -70 dBm -> CAREFUL
  quality_level= -80 dBm -> WARNING
  quality_level= -90 dBm -> CRITICAL

only 'critical' configured (careful/warning left out):
  quality_level= -50 dBm -> DEFAULT      <- a strong signal is not even OK
  quality_level= -90 dBm -> CRITICAL

only 'careful' configured:
  quality_level= -50 dBm -> DEFAULT
  quality_level= -70 dBm -> DEFAULT      <- past the careful threshold, no alert

For comparison, the base class handles partial thresholds:
  sensors base get_alert(95) with only 'critical' -> CRITICAL
```

A user who sets just `wifi_careful` in `glances.conf` gets no wifi alert at all, silently.

## The fix

Each level is read once and guarded on its own, the way `GlancesPluginModel.get_alert` already does (`plugin/model.py:854-862`) — and the way the sensors plugin is already tested for (`test_partial_type_thresholds_are_used`, `test_partial_sensor_thresholds_are_used`).

Two details worth calling out:

- **`is not None`, not truthiness.** The base class uses `if critical and value >= critical`. These limits are negative dBm values, so a `0` threshold would be discarded by a truthiness test; `is not None` is exact.
- **The `TypeError` handler is kept but narrowed.** It was added for issue #1373 (a non-numeric signal level). With the `None` comparisons gone that is now the only way it can fire, so it still does its original job and nothing else. `KeyError` is dropped because `get_limit` returns `None` rather than raising.

After:

```
only 'careful' configured:
  quality_level= -50 dBm -> OK
  quality_level= -70 dBm -> CAREFUL
```

## Tests

New `tests/test_plugin_wifi.py`, 13 cases:

- all four levels with a full config;
- each level alone still alerts (`careful` only, `warning` only, `critical` only);
- a strong signal is `OK` under any partial config, not undecorated;
- the most severe matching level wins;
- no thresholds at all stays `DEFAULT`;
- a non-numeric level stays `DEFAULT` — the #1373 guard.

**5 of the 13 fail on `develop`** and all 13 pass with the patch.

```
$ pytest tests/test_plugin_wifi.py -q
13 passed
```

### DIFF
--- a/glances/plugins/wifi/__init__.py
+++ b/glances/plugins/wifi/__init__.py
@@ -143,16 +143,26 @@ class WifiPlugin(GlancesPluginModel):
 
         :returns: string -- Signal alert
         """
-        ret = 'OK'
+        critical = self.get_limit('critical', stat_name=self.plugin_name)
+        warning = self.get_limit('warning', stat_name=self.plugin_name)
+        careful = self.get_limit('careful', stat_name=self.plugin_name)
+
+        # Each threshold is tested on its own. Chaining them meant the first
+        # undefined level raised TypeError against None and dropped the whole
+        # result to DEFAULT, so a config setting only 'careful' lost its alert
+        # entirely. GlancesPluginModel.get_alert already guards each level this
+        # way; `is not None` is used here because these limits are negative dBm
+        # values, where a plain truthiness test would discard a 0 threshold.
+        ret = 'DEFAULT' if critical is None and warning is None and careful is None else 'OK'
         try:
-            if value <= self.get_limit('critical', stat_name=self.plugin_name):
+            if critical is not None and value <= critical:
                 ret = 'CRITICAL'
-            elif value <= self.get_limit('warning', stat_name=self.plugin_name):
+            elif warning is not None and value <= warning:
                 ret = 'WARNING'
-            elif value <= self.get_limit('careful', stat_name=self.plugin_name):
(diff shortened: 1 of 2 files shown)


nicolargo:
Thanks @ntdatt812 

PR #3692, opened 2026-08-26
fix(programs): sum a program's io_counters instead of concatenating them

ntdat812:
## The bug

`io_counters` is a fixed five-slot list, not a mapping — `glances/processes.py`
documents it as:

```
[read_bytes, write_bytes, read_bytes_old, write_bytes_old, io_tag]
```

`update_program_dict` merged it with `+=`, which **concatenates** lists rather than adding
them. `cpu_times` and `memory_info` next to it are dicts and go through `sum_field_dict`;
`io_counters` had no equivalent.

Three things follow from that one operator.

**1. A program reports only its first process's disk IO.** Every reader indexes the list
(`p['io_counters'][0]`, `[2]`, `[4]` in `processlist`), so slots 0–4 keep coming from
whichever process seeded the program and the rest are never looked at:

```
process 1 io_counters: [100, 200, 10, 20, 1]
process 2 io_counters: [1000, 2000, 100, 200, 1]
program   io_counters: [100, 200, 10, 20, 1, 1000, 2000, 100, 200, 1]

displayed R/s = (100 - 10) / 1.0 = 90.0     <- shown
correct   R/s = (1100 - 110) / 1.0 = 990.0  <- actual
```

**2. The aggregation writes back into the process list.** `create_program_dict` stored the
process's own list object, and `+=` on a list extends in place, so the program and the
process share it:

```
prog['io_counters'] is procs[0]['io_counters']  ->  True
process 1 io_counters after aggregation: [100, 200, 10, 20, 1, 1000, 2000, 100, 200, 1]
```

**3. It grows without bound.** A re-sort re-aggregates the same process dicts — the same
objects in a new list — and appends again each time. Driving it through
`GlancesProcesses.get_list(sorted=True, as_programs=True)`:

```
re-sort 0: len(io_counters) =  10
re-sort 1: len(io_counters) =  15
re-sort 2: len(io_counters) =  25
re-sort 3: len(io_counters) =  35
re-sort 4: len(io_counters) =  45
re-sort 5: len(io_counters) =  55
```

## The fix

A `sum_io_counters` alongside the existing `sum_field_dict`, and a copy rather than a
borrow when the program is created.

The last slot needs different treatment from the other four. `processlist` displays a rate
only when `p['io_counters'][4] == 1` exactly, so **adding** the tag would push it to 2 for
any program with two processes and blank the R/s and W/s columns instead of correcting
them. It is OR-ed.

## Tests

`tests/test_program_io_counters.py`, nine cases. On `develop` at c15ab4c73, seven of them
fail:

```
FAILED test_io_counters_keep_their_five_slots
FAILED test_io_counters_are_summed_slot_by_slot
FAILED test_io_tag_is_set_when_any_process_reports_io
FAILED test_a_process_without_io_counters_is_skipped_not_fatal
FAILED test_a_first_process_without_io_counters_still_yields_a_list
FAILED test_the_program_does_not_borrow_the_process_list
FAILED test_repeated_aggregation_does_not_grow_the_counters
7 failed, 2 passed
```

and all nine pass with the change.

The two that pass on `develop` are deliberate: they pin the tag to `1` and to `0`, which
guards against "fix it by summing the tag too" rather than against the current bug. The
docstrings say so.

The existing `tests/test_programs_aggregation.py` and `tests/test_program_aggregation.py`
did not catch any of this because their fixtures use `io_counters: [0, 0, 0, 0, 0]` —
concatenating zeros is invisible at every index that gets read.

`tests/test_core.py`, `test_glances_stats.py`, `test_lazy_views.py` and both existing
aggregation files: 78 passed, 6 skipped.

## One thing I could not reproduce

Two of the tests cover `io_counters` being `None`. `create_program_dict` already had an
`or {}` fallback for that, so the intent was there — but `{}` is the wrong type for a
field every reader indexes as a list, and `{} += [...]` raises `TypeError` on the very
next merge. I changed the fallback to a list of five zeros.

To be clear about what that is: I could **not** reach `None` through the live pipeline.
`GlancesProcesses.get_io_counters` normalises it to `[0, 0, 0, 0, 0]` before the
aggregation ever sees it, and that is the only producer. So this part makes an existing
defensive branch coherent; it is not a fix for a crash I observed. Happy to drop it if you
would rather not carry the branch at all.

`make lint` and `make format` are clean on both files.

### DIFF
--- a/glances/programs.py
+++ b/glances/programs.py
@@ -43,13 +43,11 @@ NO_IO_COUNTERS = (0, 0, 0, 0, 0)
 
 
 def sum_io_counters(total, addition):
-    """Add addition into total slot by slot.
-
-    `+=` concatenates these lists rather than adding them, so a program reported only its
-    first process's disk IO - and, because the program borrowed that process's own list,
-    every aggregation extended it in place. The last slot is a flag rather than a total:
-    processlist displays a rate only when it is exactly 1, so it is OR-ed, not added.
-    """
+    """Add addition into total slot by slot."""
+    # `+=` concatenates these lists rather than adding them, so a program reported only its
+    # first process's disk IO - and, because the program borrowed that process's own list,
+    # every aggregation extended it in place. The last slot is a flag rather than a total:
+    # processlist displays a rate only when it is exactly 1, so it is OR-ed, not added.
     total = total or NO_IO_COUNTERS
     addition = addition or NO_IO_COUNTERS
     summed = [a + b for a, b in zip(total[:IO_COUNTERS_TAG], addition[:IO_COUNTERS_TAG])]
--- a/glances/programs.py
+++ b/glances/programs.py
@@ -22,7 +22,8 @@ def create_program_dict(p):
         'memory_percent': p['memory_percent'] or 0,
         'cpu_times': p['cpu_times'] or {},
         'memory_info': p['memory_info'] or {},
-        'io_counters': p['io_counters'] or {},
+        # A copy: the program must not borrow - and then grow - the process's own list.
+        'io_counters': list(p['io_counters'] or NO_IO_COUNTERS),
         'childrens': [p['pid']],
         # Others keys are not used
         # but should be set to be compliant with the existing process_list
@@ -33,6 +34,26 @@ def create_program_dict(p):
         'nice': p['nice'],
         'status': p['status'],
     }
+
+
+# io_counters is a fixed list, not a mapping:
+# [read_bytes, write_bytes, read_bytes_old, write_bytes_old, io_tag]
+IO_COUNTERS_TAG = 4
+NO_IO_COUNTERS = (0, 0, 0, 0, 0)
+
+
+def sum_io_counters(total, addition):
+    """Add addition into total slot by slot.
+
+    `+=` concatenates these lists rather than adding them, so a program reported only its
+    first process's disk IO - and, because the program borrowed that process's own list,
+    every aggregation extended it in place. The last slot is a flag rather than a total:
+    processlist displays a rate only when it is exactly 1, so it is OR-ed, not added.
+    """
+    total = total or NO_IO_COUNTERS
+    addition = addition or NO_IO_COUNTERS
+    summed = [a + b for a, b in zip(total[:IO_COUNTERS_TAG], addition[:IO_COUNTERS_TAG])]
+    return summed + [max(total[IO_COUNTERS_TAG], addition[IO_COUNTERS_TAG])]
 
 
 def sum_field_dict(total, addition):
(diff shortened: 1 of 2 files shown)


ntdat812:
Pushed b362582 for Codacy. It was at **34 new issues**; it is now at **12**, and I want to be
explicit about what I fixed and what I did not, rather than leave a red check unexplained.

**Fixed (23 of them).** All Prospector/pydocstyle: `D205`, `D209`, `D213`, `D400`, `D403`,
`D415` — my multi-line docstrings put a description directly under the summary line. Every
docstring is now a one-line summary and the reasoning moved into `#` comments below it. Same
words, same place, no behaviour change. Worth noting for anyone else hitting this: `ruff check`
and `ruff format` both pass clean on the file Codacy gave 23 issues to, so `make lint` does not
catch this class.

**Not fixed (12 of them).** Every one is Bandit `B101`, "use of assert detected", on an
`assert` line in a pytest file:

```
Bandit  tests/test_program_io_counters.py:35   Use of assert detected...
Bandit  tests/test_program_io_counters.py:50   Use of assert detected...
...  (12 total, all B101)
```

I have not touched these, because `assert` *is* the assertion mechanism in a pytest file and
every pytest-style test in `tests/` is written the same way — `test_programs_aggregation.py`
has 11, `test_program_aggregation.py` has 10. Rewriting them as `unittest` methods to satisfy
the check would make this file the odd one out.

There is a direct precedent, and it is the closest possible one: commit `4181b2574`
("fix: keep zero-valued fields when aggregating processes into programs") added
`tests/test_programs_aggregation.py` — the sibling test file for this very module — and its
Codacy check reads:

```
action_required :: 12 new issues (0 max.)
```

The same verdict, the same count, the same cause. It is in `develop` today.

So: happy to do whatever you prefer here. If you would rather this file used `unittest`
assertions to keep Codacy green, say so and I will convert it. If B101-on-tests is something
you already override, then nothing is outstanding from my side.

---

Separately, the full-suite attribution for this branch, since I only quoted the targeted run in
the description. Locally on Windows with the fix applied: **52 failed, 621 passed, 106 skipped**.
Excluding `test_xmlrpc.py`, `test_webui.py`, `test_browser_tui.py`, `test_browser_restful.py`
and `test_restful.py` leaves **37 failed** — and those 37 are the same family, every one a
`FileNotFoundError: [WinError 2]` from a server subprocess that will not spawn, or the
`ConnectionRefused` cascade behind it. None of them mention `programs`, `programlist` or
`io_counters`. AppVeyor is red on every open pull request in the repository for the same reason,
dependabot's included.

nicolargo:
Thanks for the patch  @ntdat812 

Just merged into develop.

PR #3694, opened 2026-08-27
fix(network): a rate of 0 in one direction must not mute the other

ntdatt812:
## The bug

`network`'s `update_views` skipped the whole interface when either direction measured 0 bytes/s:

```python
# Skip alert if no timespan to measure
if not i.get('bytes_recv_rate_per_sec') or not i.get('bytes_sent_rate_per_sec'):
    continue
```

The guard is `or`, so **one idle direction suppressed the alert for both**:

- a send-only interface saturating its uplink goes undecorated because its **rx** rate happens to be exactly 0;
- a receive-only interface — a monitor/SPAN port, a multicast receiver — the same way on **tx**;
- a fully idle interface never reaches `get_alert` at all, so its bitrate cells read `DEFAULT` instead of `OK`.

0 bytes/s is a real measurement, not a missing one.

## The fix

The rates default to `0` instead of skipping. This is what the **diskio** plugin already does — it dropped this same guard in #3684 and reads the rate through `or 0` for exactly this reason, so the two plugins now agree again.

The WebUI needs no change: `plugin-network.vue` renders `getDecoration(...)` straight from the server-supplied view, so it inherits the fix.

## Tests

Four cases in `TestNetworkPluginZeroRateAlerts`, driving the real `update_views` and reading the rendered decoration:

| case | before | after |
|---|---|---|
| saturated tx, idle rx | `DEFAULT` ❌ | `CRITICAL` ✅ |
| saturated rx, idle tx | `DEFAULT` ❌ | `CRITICAL` ✅ |
| both directions idle | `DEFAULT` ❌ | `OK` ✅ |
| both directions busy | `CRITICAL` ✅ | `CRITICAL` ✅ |

The fourth is deliberately included and passes **before and after** — it is the path that already worked, and it keeps this change honest about what it does not alter.

`tests/test_plugin_network.py`: **39 passed**.

## Note on the full local suite

Running the whole suite on this Windows host gives 53 failures, all confined to `test_restful.py`, `test_xmlrpc.py`, `test_mcp.py`, `test_perf.py` and `test_memoryleak.py` — the server/perf ones fail because `shlex.split` strips the backslashes out of the interpreter path (`D:ProjectsPersonalglances.venvScriptspython.exe`). A clean `develop` on the same host gives 52 in the same five files. **No plugin test fails on either side**, and the ±1 sits in the timing-sensitive perf/leak pair. Please read CI as the authority over my local run.

### DIFF
--- a/glances/plugins/network/__init__.py
+++ b/glances/plugins/network/__init__.py
@@ -210,13 +210,16 @@ class NetworkPlugin(GlancesPluginModel):
         for i in self.get_raw():
             if_real_name = i['interface_name'].split(':')[0]
 
-            # Skip alert if no timespan to measure
-            if not i.get('bytes_recv_rate_per_sec') or not i.get('bytes_sent_rate_per_sec'):
-                continue
-
-            # Convert rate to bps (to be able to compare to interface speed)
-            bps_rx = int(i['bytes_recv_rate_per_sec'] * 8)
-            bps_tx = int(i['bytes_sent_rate_per_sec'] * 8)
+            # Convert rate to bps (to be able to compare to interface speed).
+            # A missing rate defaults to 0 rather than skipping the interface:
+            # the guard here was `or`, so one idle direction suppressed the
+            # alert for BOTH. A send-only interface saturating its uplink went
+            # undecorated because its rx rate happened to be exactly 0, and a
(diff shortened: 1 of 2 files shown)


nicolargo:
Thanks for the PR @ntdatt812 

Merged into develop.

PR #3699, opened 2026-08-27
fix(webui): colour the process MEM% cell by memory_percent, not cpu_percent

ntdatt812:
## The bug

`plugin-processlist.vue` decorates the **MEM%** column using the **CPU%** value:

```js
getMemoryPercentAlert(process) {
    return GlancesHelper.getAlert("processlist", "processlist_mem_", process.cpu_percent);
}
```

The cell right next to it renders `process.memory_percent`, so in the most-looked-at table in Glances the number and its colour come from different fields. Both render paths — the flat list and the tree view — go through this one helper.

The curses UI gets it right:

```python
def _get_process_curses_memory_percent(self, p, selected, args):
    msg = self.layout_stat['mem'].format(p['memory_percent'])
    alert = self.get_alert(p['memory_percent'], ..., header="mem")
```

## What a user sees, with the shipped defaults

`conf/glances.conf` ships `mem_careful/warning/critical = 50/70/90`. Running the real `getAlert` from `services.js` against them:

```
a memory hog, idle CPU (cpu 0.3%, mem 94.2%)
   MEM% cell BEFORE: ok           <- 94% of RAM, no alert
   MEM% cell AFTER : critical

a busy loop, tiny RSS (cpu 99%, mem 0.4%)
   MEM% cell BEFORE: critical     <- red on a 0.4% footprint
   MEM% cell AFTER : ok
```

The process you'd open Glances to find — the one eating memory — is exactly the one that stays green, while a CPU-busy process with a negligible footprint is flagged instead.

## Bundle

`public/glances.js` is rebuilt. A byte-offset diff is useless here because `memory_percent` is 3 bytes longer than `cpu_percent`, so everything after it shifts. Instead, undoing just that identifier in the rebuilt bundle:

```js
const normalized = rebuilt.replace('processlist_mem_",t.memory_percent',
                                   'processlist_mem_",t.cpu_percent');
normalized === committed   // true
```

Byte-identical. So the rebuild carried nothing but this change, and the size delta is exactly +3.

## Related

Same class of defect as #3698, which I opened for `plugin-percpu.vue` (the iowait cell coloured by `system`). They're independent files, so I kept them as separate PRs — merge order doesn't matter.

I checked every other `GlancesHelper.getAlert` call site in `js/components/`; these two were the only mismatches. `plugin-raid.vue` already mirrors `RaidPlugin.raid_alert` correctly.

### DIFF
--- a/glances/outputs/static/js/components/plugin-processlist.vue
+++ b/glances/outputs/static/js/components/plugin-processlist.vue
@@ -702,7 +702,7 @@ export default {
 			return GlancesHelper.getAlert(
 				"processlist",
 				"processlist_mem_",
-				process.cpu_percent,
+				process.memory_percent,
 			);
 		},
 		getDisableStats() {


nicolargo:
Hi @ntdatt812 

all the lastest WebUI PR generate conflict on the glances.js file.

This file is generate dynamically during release. It should not be committed in PR.  

Can you rework all previous PR ? You can also put all the PR in one big PR.

Nicolas

ntdatt812:
Done, and thanks for saying it plainly. All five open WebUI PRs are rebased onto current `develop` and no longer touch `glances/outputs/static/public/` at all:

| PR | file | diff |
| --- | --- | --- |
| #3695 | `plugin-ports.vue` | +7 / -7 |
| #3696 | `plugin-amps.vue` | +10 / -12 |
| #3697 | `plugin-vms.vue` | +15 / -1 |
| #3698 | `plugin-percpu.vue` | +2 / -2 |
| #3699 | `plugin-processlist.vue` | +1 / -1 |

One source file each, nothing generated. I rebuilt each branch from `develop` and re-applied only the `.vue` change, then checked the source diff hashes against the old tips to be sure the rework did not quietly alter a fix: all five are byte-identical to what you already reviewed. The bundle conflicts are gone with it, since that file is the only thing they collided on.

I saw 076469f3 land while I was doing this. Understood, and I have `git checkout -- glances/outputs/static/public/` in my loop now. Sorry for the noise: I had been committing the rebuild so the change was testable straight from the branch, which was the wrong trade.

On merging them into one PR: happy either way. They are five unrelated one-line-class fixes in five different plugins, all of the same shape, each reading the wrong field for a colour, so as separate PRs you can drop any one of them without touching the rest. But if one PR is less work for you, say so and I will fold them into a single branch today.

nicolargo:
Perfect @ntdatt812 

Merged into develop, i will make some smoke tests on my side before generating the glances.js file.

Thanks for your work !

Issue #1528, opened 2019-08-08
Error with IP Plugin : object has no attribute


in2out:
I have a problem (centos 7.2)

xxxxx# glances
Error while initializing the ip plugin ('NoneType' object has no attribute 'split')

xxxxx# glances  -V
Glances v3.1.1 with psutil v5.6.3

it cause below part.

plugins/glances_ip.py
return  ', '.join(set([x.strip() for x in ip.split(',')]))

ip is none...

nicolargo:
Corrected by last commit.

THanks for the issue report.

Globegitter:
@nicolargo I am on Ubuntu 19.10 with python 3.7 and getting the issue with glances 3.1.2:

```
Error while initializing the ip plugin ('NoneType' object has no attribute 'split')
Exception ignored in: <module 'threading' from '/usr/lib/python3.7/threading.py'>
Traceback (most recent call last):
  File "/usr/lib/python3.7/threading.py", line 1307, in _shutdown
    lock.acquire()
  File "/home/markus/.local/lib/python3.7/site-packages/glances/__init__.py", line 68, in __signal_handler
    end()
  File "/home/markus/.local/lib/python3.7/site-packages/glances/__init__.py", line 83, in end
    sys.exit(0)
SystemExit: 0
```

Any idea why the issue could still show up? Or how to work around that.

Globegitter:
Ah yes, it works fine without the plugin so at least I can start up glances :)

NickWaterton:
Ubuntu 18.04, python 3.6, Glances 3.1.3 psutil 5.6.7 has the same issue. Starts with ip plugin disabled.

nicolargo:
The DEVELOP branch has a new patch.

It solve this second issue concerning the IP plugin initialisation .

maxnbk:
experienced this issue, patched 3.1.3 (with psutil 5.6.7) with commit and works fine.

davidmaxwaterman:
I'm seeing this with glances v3.1.3-1 on Ubuntu Ubuntu 20.04.1 LTS. Any idea how to fix it?

davidmaxwaterman:
Still getting this. Any chance it could be reopened?
```
10:59:32 ~$ glances
Error while initializing the ip plugin ('NoneType' object has no attribute 'split')
^C^CException ignored in: <module 'threading' from '/usr/lib/python3.8/threading.py'>
Traceback (most recent call last):
[...]

PR #3702, opened 2026-08-30
IP plugin: prefer the default-route address over interface order

jamescowens:
## The problem

`get_ip_address()` in `glances/globals.py` returns the address of the **first up,
non-loopback interface** psutil enumerates. On any host running Docker, VMware,
libvirt or similar, a virtual bridge (`docker0`, `vmnet8`, `br-*`, ...) can occupy
that position and silently win, so the IP plugin shows a bridge address (for
example `172.17.0.1`) as the host's primary IP. Reported by users in #3465.

The nasty part is that the result depends on **interface creation order**, so a
host can display the correct address for months and regress after something as
routine as a NetworkManager restart recreating a bond interface. On one of my
hosts (22 interfaces: a bond, 3 docker bridges, 2 VMware vmnets, 10 veths) the
correct answer currently survives only because the bond happens to enumerate
before six other candidates that would all "win" if ordered first.

I believe this regressed when the `netifaces` dependency (which provided
`gateways()`) was dropped — the same gap discussed in #3073.

## The fix

Ask the OS routing table which source address it would use for default-route
traffic: `connect()` on a `SOCK_DGRAM` socket **sends no packet** — it only
resolves the route — and honours metrics, policy routing and per-route source
hints, identically on every platform Glances supports (no `/proc` parsing, no
new dependency). The probe address is from the documentation range
(RFC 5737 / RFC 3849) and is never actually contacted.

The previous interface scan is kept as the fallback for hosts without a default
route, and the pre-existing tests now pin exactly that fallback behaviour. A
`None` netmask in the (unlikely) case where the routed address is not listed by
psutil is already handled by `ip_to_cidr()` (#1528). A loopback/unspecified
probe result (possible on hosts that locally blackhole documentation/bogon
ranges) is rejected and falls back to the scan, so a loopback address can
never be reported.

Two design notes for review:

- **Why not reuse `get_default_gateway()`?** It parses `/proc/net/route`, so it
  is Linux-only, and it returns the *gateway's* address rather than the local
  source address the plugin needs. The socket probe works on every supported
  platform and delegates the choice to the same code path real traffic uses.
- **Second consumer**: `servers_list_dynamic.py` uses `get_ip_address()[0]` as
  the zeroconf bind address. The change improves that case too — previously a
  virtual bridge address could be announced to LAN peers as the server address.

## Verification

- `pytest tests/test_globals.py` — 5 passed (2 new: default-route address wins
  over interface order; routed address without a psutil entry yields a `None`
  netmask. 3 pre-existing, now exercising the no-default-route fallback).
- `ruff check` / `ruff format --check` clean on both changed files.
- Live on the 22-interface host above: returns the bond address
  (`192.168.0.x`) for IPv4 and the global address for IPv6, independent of
  enumeration order. The old walk's order-dependence is reproduced in the new
  unit test, where `docker0` enumerating first used to win.

Fixes the behaviour discussed in #3465; related to #3073.

### DIFF
--- a/glances/globals.py
+++ b/glances/globals.py
@@ -17,6 +17,7 @@ import base64
 import errno
 import functools
 import importlib
+import ipaddress
 import multiprocessing
 import os
 import platform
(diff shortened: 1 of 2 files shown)


nicolargo:
Thanks for the PR @jamescowens 

Is this code compliant with non Linux Operating System ?

jamescowens:
Yes — portability was a design constraint, and I've now verified it empirically on every platform Glances supports except SunOS:

| platform | IPv4 probe | IPv6 probe |
|---|---|---|
| Linux (3 hosts, one with 22 interfaces) | correct LAN address | correct global address |
| FreeBSD 15.0 | correct | correct global |
| OpenBSD 7.8 | correct | no v6 route → falls back cleanly |
| macOS (Darwin 23.6) | correct (VM NAT address — its real routed interface) | correct |
| Windows 11 (native Python) | correct | no v6 route → falls back cleanly |
| Windows 11 / WSL2 | correct (WSL NAT — its real routed interface) | no route → falls back cleanly |

The mechanism is plain BSD-sockets API — a connected `SOCK_DGRAM` socket plus `getsockname()` — which is native on the BSDs, POSIX on Linux/macOS, and standard Winsock behaviour on Windows (implicit bind on `connect()`, confirmed above). No `/proc` parsing, no new dependency; that portability is why I didn't build on `get_default_gateway()`, which is Linux-only and returns the gateway's address rather than the local source address the plugin needs.

Equally important: the rows above where a host *lacks* a route demonstrate the designed degradation on real systems — every failure path (no default route, family unsupported, filtered probe, loopback result) falls through to the previous interface scan, kept verbatim. On any platform where the probe can't work, behaviour is identical to the current release.

The new unit tests mock the socket, so they're deterministic on your full CI matrix. On the checks attached to this PR: the AppVeyor failure is pre-existing — the last dozen builds fail identically on every branch including merged PRs (the harness can't launch the server subprocess before any of this code runs).

Happy to add a platform gate if you'd prefer the old behaviour kept outside Linux, but as the table shows, the socket path is the portable part of the change.

nicolargo:
Thanks for the issue report and the patch @jamescowens !

Just reviewed and merged into develop.

jamescowens:
You're very welcome! :)

Issue #3299, opened 2025-10-06
[UI] By default hide the VIRT memory (display only RSS)


nicolargo:
For the moment both VIRT/RSS are displayed by default.

Change this behavior to only display RSS.

Additionally: Make the list configurable (already done for memory but not at VIRT/RSS granularity).

nicolargo:
```
[processlist]
# Disable display of virtual memory
disable_virtual_memory=True
```

Result:

<img width="803" height="422" alt="Image" src="https://github.com/user-attachments/assets/90047b9b-87e4-46eb-8245-c75f9f863731" />

nicolargo:
Done on the develop branch, will be released in target milestone.

PR #3729, opened 2026-09-11
Read a False config switch as False in get_conf_value

datrixlab:
#### Description

`docs/aoa/ps.rst` documents `disable_virtual_memory` as a switch in `[processlist]`, but
setting it to `False` **hides** the VIRT column in the curses UI, and a numeric value
raises.

`get_conf_value(convert_bool=True)` (`glances/plugins/plugin/model.py`) returned:

```python
return bool(ret[0]) if convert_bool else ret
```

`load_limits` stores a value that is not a number as a one-item list of strings, so
`disable_virtual_memory=False` reaches this as `['False']`, and `bool('False')` is `True`.
A value that *is* a number is stored as a float, which has no `[0]`.

Against `develop`, building `ProcesslistPlugin` from a config with only that key:

```
disable_virtual_memory=True  -> True
disable_virtual_memory=False -> True        # should be False
disable_virtual_memory=0     -> TypeError: 'float' object is not subscriptable
disable_virtual_memory=1     -> TypeError: 'float' object is not subscriptable
```

The two other readers of boolean limits already do it right:

- the WebUI, for this very key (`plugin-processlist.vue`, `getDisableVms`):
  `return ret[0].toLowerCase() === "true" ? true : false;`
- the `*_log` check a few lines above in `model.py`:
  `return self._limits[stat_name + '_log'][0].lower() == 'true'`

so the same `glances.conf` currently shows VIRT in the WebUI and hides it in the curses UI.

The fix reads the text the same way and takes `bool()` of a float. The only callers of
`convert_bool=True` are the two `disable_virtual_memory` checks in
`plugins/processlist/__init__.py`.

#### Resume

* Bug fix: yes
* New feature: no
* Fixed tickets: none open — the option was added in #3299

#### Tests

`test_708_conf_value_convert_bool` in `tests/test_core.py` covers `True`, `true`, `False`,
`1`, `0`, and an unset key falling back to the default. Reverting `model.py` alone fails it
at the `False` case (`AssertionError: True is not False`).

`tests/test_core.py`: 49 passed, 6 skipped. `ruff check` clean; `ruff format --check`
reports only the existing `tests/test_core.py:422`, present on an unmodified `develop`.

### DIFF
--- a/glances/plugins/plugin/model.py
+++ b/glances/plugins/plugin/model.py
@@ -1024,9 +1024,14 @@ class GlancesPluginModel:
 
         try:
             ret = self._limits[plugin_name + '_' + value]
-            return bool(ret[0]) if convert_bool else ret
         except KeyError:
             return default
+        if not convert_bool:
+            return ret
+        # load_limits stores a key as a one-item list of strings, or as a float when
+        # it parses as a number. bool('False') is True and a float has no [0], so
+        # read the text the way the *_log check above and the WebUI already do.
+        return ret[0].lower() == 'true' if isinstance(ret, list) else bool(ret)
 
     def is_show(self, value, header=""):
         """Return True if the value is in the show configuration list.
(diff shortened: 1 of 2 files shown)


nicolargo:
Thanks again @datrixlab 

Merged into develop.

PR #3732, opened 2026-09-12
Remember that a connections probe was disabled

datrixlab:
## Description

The connections plugin stops using a probe that fails — `psutil.net_connections()` raising, or the `nf_conntrack` files not being present — and records that by writing `net_connections_enabled` / `nf_conntrack_enabled` into `stats`.

But `update()` starts every refresh with `stats = self.get_init_value()`, which is `copy.copy(self.stats_init_value)`, and `stats_init_value` has both flags `True` (`glances/plugins/connections/__init__.py:98`). The flag is written into a dict that is thrown away at the end of the refresh, and the next refresh calls the failing probe again.

Measured by pointing `conntrack` at a file that does not exist and refreshing three times:

```
cycle 0: stats['nf_conntrack_enabled'] = False
cycle 1: stats['nf_conntrack_enabled'] = False
cycle 2: stats['nf_conntrack_enabled'] = False
conntrack read attempts: 3          (1 after this PR)
"Can not get network connections track" warnings: 3   (1 after this PR)
```

So on any host without `/proc/sys/net/netfilter/nf_conntrack_*` — macOS, Windows, a Linux kernel without the module — the warning plus its `Disable connections track` info line are logged on every single refresh.

The same applies to the `psutil.net_connections()` guard, and that one matters beyond the log: the plugin is disabled by default *"because it consumes lots of CPU"*, so a user who turns it on (config or the `K` key) on a host where the call raises, such as macOS without root, pays for the failing call every refresh.

Every other plugin that gives up keeps the flag outside the per-refresh stats: `mem` with `self.zfs_enabled` (`mem/__init__.py:134`, read at `:180`), `sensors` with its grabber's `init` flag, `wifi` with a module-level check.

## Resume

- Keep `net_connections_enabled` / `nf_conntrack_enabled` on the plugin instance, set to `False` the first time a probe fails.
- Copy them into `stats` at the start of each refresh, so `update_views()`, `msg_curse()` and the REST/WebUI payload still carry the same two fields as before.

Tests in `tests/test_connections_states.py`:

- a failed conntrack read happens once across three refreshes, and the flag stays `False`
- a raising `psutil.net_connections` is called once across three refreshes
- guard: when both probes work, they keep running every refresh and `nf_conntrack_percent` is still computed

The tests drive the real `update()` with the refresh timer pushed past its deadline between calls, the way the main loop reaches it; without that, the refresh timer alone would make any version of the code look correct.

With `glances/plugins/connections/__init__.py` reverted, the first two fail (3 reads / 3 calls instead of 1) and the guard passes. `pytest tests/test_connections_states.py tests/test_core.py tests/test_plugin_ports.py` → 75 passed, 6 skipped. `ruff check` and `ruff format --check` clean.

## Fixed tickets

None open; I did not find an issue for this.

### DIFF
--- a/glances/plugins/connections/__init__.py
+++ b/glances/plugins/connections/__init__.py
@@ -102,12 +102,20 @@ class ConnectionsPlugin(GlancesPluginModel):
         # We want to display the stat in the curse interface
         self.display_curse = True
 
+        # A probe that fails is not tried again for the rest of the session. These
+        # have to live on the plugin: update() starts every refresh from a fresh copy
+        # of stats_init_value, so a flag written into stats was forgotten immediately
+        # and the call was made (and the warning logged) again on the next refresh.
+        self.net_connections_enabled = True
+        self.nf_conntrack_enabled = True
+
     def update_for_net_connections_method(self, stats):
         try:
             net_connections = psutil.net_connections(kind="tcp")
         except Exception as e:
             logger.warning(f'Can not get network connections stats ({e})')
             logger.info('Disable connections stats')
+            self.net_connections_enabled = False
             stats['net_connections_enabled'] = False
 
             return stats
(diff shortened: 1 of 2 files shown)


nicolargo:
Thanks @datrixlab 

Merged into develop.

Issue #3643, opened 2026-08-05
Make Glances compliant with MCP lib >= 2


nicolargo:
Related to the issue https://github.com/nicolargo/glances/issues/3640.

Glances is not compliant with MCP lib >= 2.

Migration guide should be follow; https://py.sdk.modelcontextprotocol.io/migration/

nicolargo:
Note: no rush...

_**The v1.x maintenance line keeps receiving critical bug fixes and security patches, and its documentation is at [/v1/](https://py.sdk.modelcontextprotocol.io/v1/). If your package depends on mcp, keep a <2 upper bound until you've migrated.**_

Issue #3640, opened 2026-08-04
mcp extra allows mcp 2.0.0 which breaks --enable-mcp (No module named 'mcp.server.fastmcp')


zeusecurity:
**Glances version:** 4.5.6 (pip/pipx install)
**Python:** 3.12 (Ubuntu 24.04 pipx venv)

The `mcp` optional dependency is declared as `mcp>=1.0.0; extra == "mcp"`, which allows the recently released **mcp 2.0.0**. The 2.0 SDK removed/moved `mcp.server.fastmcp`, so `glances -w --enable-mcp` fails at startup even though the `mcp` package is installed:

```
ModuleNotFoundError: No module named 'mcp.server.fastmcp'
...
RuntimeError: The 'mcp' package is required for MCP support. Install it with: pip install 'glances[mcp]'
```

(the RuntimeError message is misleading here — the package IS installed, just incompatible)

`glances/outputs/glances_mcp.py` imports:
```python
from mcp.server.fastmcp import FastMCP
from mcp.server.sse import TransportSecuritySettings
```
both of which are gone in mcp 2.x.

**Workaround:** `pip install 'mcp<2'` inside the venv fixes it.

**Suggested fix:** pin the extra to `mcp>=1.0.0,<2` (or migrate to the 2.x API).

nicolargo:
Thanks for the head-up @zeusecurity !

Will pin it in the develop branch.

nicolargo:
Pin done in the develop branch.

Migration path followed by another ticket: https://github.com/nicolargo/glances/issues/3643

PR #3735, opened 2026-09-14
feat(mcp): support MCP Python SDK 2.x (#3643)

anandghegde:
#### Description

Following the [v1 to v2 migration guide](https://py.sdk.modelcontextprotocol.io/migration/), two changes in MCP Python SDK 2 affect `glances/outputs/glances_mcp.py`:

- `FastMCP` was renamed to `MCPServer` and `mcp.server.fastmcp` is gone. That import error is why `--enable-mcp` did nothing with mcp 2.0 (#3640).
- `transport_security` moved from the constructor to `sse_app()`. If it's left in the constructor, the DNS rebinding allow-list built from `mcp_allowed_hosts` is no longer applied.

The module now imports `MCPServer` and falls back to `FastMCP` on 1.x. It passes `transport_security` to `sse_app()` on 2.x and to the constructor on 1.x. `TransportSecuritySettings` is imported from `mcp.server.transport_security`, where the guide documents it. That module exists in 1.x too. Nothing else in the resource and prompt code needed to change.

The `mcp` extra now allows `mcp>=1.0.0,<3`. I kept 1.x working instead of requiring 2.x because semgrep in the dev group pins `mcp==1.29.0`. With `mcp>=2,<3`, `uv lock` still succeeds, but it quietly drops semgrep to 1.79.0 (from 2024). With `<3`, the lock keeps semgrep 1.177.0 and mcp 1.29.0, so the generated requirements files stay valid and don't need regenerating. `pip install glances[mcp]` now gets mcp 2.x. Once semgrep allows mcp 2, the lower bound can go up and the 1.x branch can be removed.

`tests/test_mcp.py` passes plain strings to `read_resource()`, since v2 only accepts `str`. It reads `uriTemplate` through `model_dump(by_alias=True)` because v2 renamed the attributes to snake_case. With those changes the same tests run on both SDKs.

How I tested (macOS, Python 3.12/3.13, uv venvs):

- `pytest tests/test_mcp.py`: 25 passed with mcp 2.2.0, and 25 passed with mcp 1.30.0. On unmodified `develop` with mcp 2.2.0 the suite fails, because the server never mounts `/mcp`.
- `uv pip install -e '.[web,mcp]'` in a clean venv resolves mcp 2.2.0. `pytest tests/test_mcp.py tests/test_core.py` then gives 76 passed, 4 skipped.
- Full `pytest tests` with mcp 2.2.0, skipping memoryleak, perf and webui: 929 passed. The 15 failures are all in `test_actions_sanitize.py` and fail the same way on unmodified `develop` in that venv, which doesn't have the `action` extra (chevron).
- Ran `glances -w --disable-webui --enable-mcp` with mcp 2.2.0 and connected with an mcp 2.2.0 SSE client:

```
server: Glances protocol: 2025-11-25
tools: []
resources: ['glances://plugins', 'glances://stats', 'glances://limits']
templates: ['glances://stats/{plugin}', 'glances://stats/{plugin}/history', 'glances://limits/{plugin}']
prompts: ['system_health_summary', 'alert_analysis', 'top_processes_report', 'storage_health']
glances://stats/cpu keys: ['cpucore', 'ctx_switches', 'idle', 'interrupts', 'nice', 'soft_interrupts', 'syscalls', 'system'] total = 18.2
prompt top_processes_report: You are a system administrator assistant. Analyze the following top 3 processes sorted by  ...
unknown plugin -> MCPError Error creating resource from template glances://stats/nope
```

- DNS rebinding protection still works on 2.x: `curl -H "Host: evil.example" http://127.0.0.1:PORT/mcp/sse` returns 421, and `localhost` returns 200. The server logs a `Request validation failed` traceback for the rejected request, but mcp 1.30 logs the same thing.
- `ruff check` and `ruff format --check` (0.15.4, the pre-commit version) pass on the changed files.

One behaviour change comes from the SDK itself: in 2.x, sync resource and prompt handlers run on a worker thread instead of the event loop. They only read from the stats object, and the tests and manual calls above ran without problems.

#### Resume

* Bug fix: no
* New feature: yes
* Fixed tickets: #3643


nicolargo:
Just merged into develop.  

Thanks for the PR @anandghegde  !

PR #3736, opened 2026-09-14
fix(stats): set `disable_<plugin>` for additional plugins so `is_enabled()` works for them

cyrbygh:
#### Description

I'm trying to add a custom plugin for reporting ZFS pool size for a pool that isn't mounted (only contains encrypted backup data sets this particular backup machine purposely doesn't have the keys for), so the built-in `fs` plugin can't see it. I added an additional plugin for calling `zpool list` directly, but curiously Glances doesn't seem to pick it up.

The issue appears to be that `load_additional_plugins()` never sets `args.disable_<plugin>`, so `is_enabled()` always returns `False` for plugins loaded via `-P`. The fallback chain in `is_enabled()` checks `args.enable_<plugin>` (absent defaults to `True`), then returns `True is False` = `False`.

This PR mirrors the same logic already present in `_load_plugin()` / `load_plugins()` for built-in plugins.

#### Resume

* Bug fix: yes
* New feature: no
* Fixed tickets: N/A


nicolargo:
Just merged into develop !

Thanks for the patch @cyrbygh  !

PR #3737, opened 2026-09-20
feat(gpu): report real Intel GPU load from fdinfo engine counters

DocMAX:
## Problem

The Intel backend reports the `gt_act_freq_mhz / gt_max_freq_mhz` ratio as `proc`. That number stays high whenever the GPU is clocked up, even with no actual work running. Observed on an Iris Xe (Raptor Lake-P): frequency pinned at 1450/1450 MHz while idle reads as **100% load**.

## Change

Sum the per-client `drm-engine-*` time counters from `/proc/*/fdinfo` (i915/xe drivers) and report the delta over the refresh interval as the processor load -- the same approach the ARM backend already uses. Details:

- new `parse_fdinfo()` / `aggregate_fdinfo()` in `glances/plugins/gpu/cards/intel.py`, with pdev matching and single-device-per-driver fallback (mirrors `arm.py`)
- `drm-engine-capacity-*` entries are counts, not time, and are excluded
- summed engine time can exceed wall-clock (parallel engines) and is clamped to 100
- when no fdinfo data is available (no GPU client, first call after startup, unreadable `/proc`), the previous frequency ratio is kept as a fallback, so `proc` is never blank
- `mem`/`temperature`/`fan_speed` remain `None` (not exposed for Intel); docs updated

## Testing

- new fixtures under `tests-data/plugins/gpu/intel/` and `TestIntel*` classes in `tests/test_plugin_gpu.py`: `55 passed`
- `ruff check` clean
- live on Intel Iris Xe (i915): **77%** during `llama-server` Vulkan inference, **0%** right after while the frequency was still pinned at max -- exactly the case the old code got wrong


nicolargo:
Hi @DocMAX 

Thanks for the PR.

I can not accept it as-it because it mix backend / frontend.

Your change in the plugin is interesting but focus on the GPU stats from fdinfo and do not change the UI in the same PR.

Also note that Glances v4 is frrezed in term of API change, so do not include any change in the fields_description.

My proposal is the following:
- rework your PR to only include GPU stats grab from fdinfo
- keep your change in the API and on the UI when Glances v5 will be release

Thanks again !

Issue #3611, opened 2026-07-13
VideoCore (v3d) memory shows ~93% on Raspberry Pi 5 with gpu_mem=4M — misleading denominator from drm-total-memory


tomaszgrabowskiop:
Hardware	Raspberry Pi 5
OS	Debian 13 (Trixie)
Kernel	6.18
Glances	4.5.5 (latest stable)
gpu_mem (firmware)	4 MB
[...]

PR #3738, opened 2026-09-20
feat(gpu): per-process GPU column and Intel VRAM usage

DocMAX:
Stacked on #3737 (contains its commit until that merges).

## Per-process GPU% column (nvtop-style, Intel i915/xe)

New `GPU%` column right next to `CPU%` in the process list, fed by per-PID deltas of the `drm-engine-*` fdinfo counters -- the same source nvtop uses.

- `v` sorts by GPU%, Tab cycles through it, `--sort-processes` accepts `gpu_percent`
- the column only appears when an Intel GPU is detected (checked every refresh, so a GPU bound away for passthrough hides it instead of showing stale zeros) and respects `disable_stats`
- top GPU consumer gets MAX highlight via `max_values`

Implementation: `get_per_pid_gpu_percent()` + `intel_gpu_present()` in `glances/plugins/gpu/cards/intel.py`; merge + max in `ProcesslistPlugin.update()`; sorting via the generic `sort_stats` path (only `sort_processes_stats_list`/`sort_for_human` extended, no core logic change).

## Intel VRAM usage

The GPU section `mem` field was hardcoded `None` for Intel. It now reports GPU-resident shared buffers: summed `drm-resident-*` region counters over CmaTotal/MemTotal (reuses the ARM helpers, so the #3611 lesson -- never `drm-total-memory` as denominator -- applies here too). `shared/active/purgeable` are subsets and excluded from the sums.

## Testing

- fixtures under `tests-data/plugins/gpu/intel/`, `TestIntelPerPid`/`TestIntelMem` classes plus new `tests/test_plugin_processlist_gpu.py` (merge, visibility rules, sorting, hotkey): `70 passed`
- `ruff check` clean
- live on Intel Iris Xe (i915): device `mem` reads **12%** (7.7 GiB resident GPU buffers, mostly `llama-server`/Vulkan, over 62.4 GiB MemTotal); per-PID `llama-server` shows up top when inferring


DocMAX:
Note on coloring: the GPU% cells use `header="gpu"`, which has no `[processlist] gpu_*_thresholds` keys, so they render `DEFAULT` (+ `?` for unknown). This is deliberate -- no invented thresholds. If you want CAREFUL/WARNING/CRITICAL + MAX highlighting like the CPU% column, say so and I will add `gpu_careful/warning/critical` defaults to the section.

DocMAX:
Update: both review points from local testing are fixed.

- **Sort loop skipped the column**: the sort order was `cpu, mem, ..., gpu(last)` while the column sits between CPU% and MEM%. `gpu_percent` is now adjacent to `cpu_percent` in `sort_processes_stats_list`, so Shift+Left/Right steps onto the visually adjacent column (test: `test_gpu_next_to_cpu_in_sort_loop`).
- **Colors**: `processlist` now defaults `gpu_careful/warning/critical` to 50/70/90 (code default in `sections_set_default`, like cpu/mem, plus documented keys in the bundled conf). Verified live: 77->WARNING, 60->CAREFUL, 10->OK -- same behavior as the CPU% column.

nicolargo:
As said in the previous PR, this one will be done when Glances v5 (it's an API change on processlist).

DocMAX:
Update: per-process GPU memory column added (same branch).

New `GMEM` column next to `GPU%`: per-PID `drm-resident-*` bytes via `get_per_pid_gpu_mem_bytes()` -- a point-in-time gauge, same visibility rules as GPU% (Intel present, not disabled). Displayed with `auto_unit` like VIRT/RES, no thresholds. Shared buffers may be counted once per holder, same limitation as nvtop's MEM column.

Live on Iris Xe: `llama-server` shows 18.2G GMEM while idle (model buffers resident, 0% load).

DocMAX:
Update: GMEM is now GMEM% -- percent, colored, sortable.

- per-process GPU memory is now a percentage of system memory (like MEM%), not raw bytes
- own `gmem_careful/warning/critical` 50/70/90 defaults (code + bundled conf) with MAX highlight, same behavior as the other columns
- sortable: `gpu_mem` sits right after `gpu_percent` in the sort loop, new `y` hotkey, `--sort-processes` choice and help entry

Live on Iris Xe: `llama-server` at 29.1% GMEM while idle.

DocMAX:
Update: both memory views now exist side by side.

- `GMEM%` next to `GPU%`: percent of system memory, colored, sortable (`y`, sort loop, `--sort-processes`)
- `GMEM` after `MEM%`: raw bytes with `auto_unit`, like VIRT/RES (no thresholds)
- each column individually hideable via `disable_stats`

